PR #21, opened 2017-05-24
ci(deploy): set the GIT_PUBLISH_URL in a different way

thewtex:
Attempt to address "{ ProcessError: ERROR: The key you are authenticating with has been marked as
read only."


jourdain:
I see you think that the variable may not be available... You can also put it in the ENV var of the settings in Circle ci.

PR #39, opened 2017-06-01
feat(config): make the location of the ImageIOs and web workers confi…

thewtex:
…gurable

With node, it can be configured as described in the node-config documentation:

  https://www.npmjs.com/package/config

With browserify, it can be configured by supplying a JSON file with something
like aliasify for 'config':

  https://github.com/benbria/aliasify

With webpack, it can also be configured by supplying a JSON file:

  https://github.com/lorenwest/node-config/wiki/Webpack-Usage

### DIFF
--- a/config/browser-test.json
+++ b/config/browser-test.json
@@ -1,0 +1,6 @@
+{
+  'itk': {
+    imageIOsPath: '/itkImageIOs',
+    webWorkersPath: '/itkWebWorkers'
+  }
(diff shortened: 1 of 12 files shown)


jourdain:
Hum not sure that will solve the issue.

thewtex:
I'm going to forge ahead. Shooting for a good demo :clapper:

PR #78, opened 2018-02-21
refactor(WebWorkers): Use webworker-promise instead of promise-worker…

thewtex:
…-transferable

More capable and better maintained.

CC: @agirault


agirault:
Looks nice! Not passing checks yet?

thewtex:
Follow-up added.

PR #80, opened 2018-02-21
docs(Build): Provide helpful error message when docker is not available

thewtex:
When running npm run commit.


thewtex:
Addresses #7

PR #87, opened 2018-03-02
fix(ImageIOWorker): Wrap the returned errors in a Promise

thewtex:
This improves how the error is handled downstream and improves the message.


thewtex:
@agirault the second patch should fix the 'out of memory' errors.

agirault:
magnificent

agirault:
@thewtex That did not trigger a new release?

thewtex:
I will shoot for a new release tomorrow.

PR #88, opened 2018-03-09
Mesh io

thewtex:


agirault:
Will that make the next release heavier? Or can we select what part of itk-js we want to package? I'm asking because I know that we have been copying the whole ITK directory in dist/ so far

thewtex:
Yes, you could limit the parts that you want to copy. If you just want the image IO support, copy `itk/ImageIOs`. For mesh IO support (there is currently only one format supported, but more are coming), copy `itk/MeshIOs`.

agirault:
nice thanks. and WebWorkers also?

thewtex:
Right, and `WebWorkers`. The rest of the files are just used at build time.

I will create an example and add you as a reviewer.

PR #97, opened 2018-03-16
feat(Examples): Add Webpack example

thewtex:
@agirault It turns out we cannot bundle the web workers because of [a bug](https://github.com/webpack-contrib/worker-loader/issues/45) in `worker-loader`.


agirault:
😢

PR #149, opened 2018-12-16
Hello world example

thewtex:
@fbudin69500 @jcfr @HastingsGreer @floryst please take a look :book:


thewtex:
Merging so doc's will be published here:

  http://insightsoftwareconsortium.github.io/itk-js/

Feedback welcome! :microphone: :speaker:

HastingsGreer:
The command 
```	npx itk-js build . ```
fails silently for me. Is there a verbose flag?

thewtex:
@HastingsGreer there is no output at all? What version of node / npm / npx do you have?

thewtex:
@HastingsGreer super -- thanks for testing and the information! We probably have an issue with the Docker installation's ability to mount local volumes... I will follow-up IRL to discuss details and see if we can detect the conditions.

PR #150, opened 2018-12-17
feat(Dependencies): Bump ITK to 2018-12-16 master

thewtex:
Addresses final VXL warnings (hooray!).

Bring in modular FreeSurfer mesh module.


thewtex:
@hjmjohnson thank you so much for the massive VXL improvements. There used to be a huge amount of warnings related to VXL. The builds are now clean! :1st_place_medal:

PR #179, opened 2019-01-31
feat(Examples): Node.js examples uses synchronous API

thewtex:


thewtex:
@yh3ds here is an example with the synchronous API.

Issue #251, opened 2019-08-23
Provide an function to list the series available in a list of DICOM files


thewtex:
A high level API to identify distinct series in a list of DICOM files. The series can then be passed to `itk/readImageLocalDICOMFileSeries` or metadata investigated with the function provided by #250 .

Issue #250, opened 2019-08-23
Provide a high level API function to list DICOM metadata in a file


thewtex:
Provide basic DICOM metadata in a convenient API.

Internally, this function could use dcm.js or daikon, for example.

PR #252, opened 2019-09-05
WIP: Dicom Parsing Example

agirault:
Early work in progress showcasing an example that parses and organizes DICOM instances before reading them with ITK.

Opening this PR mainly to enable discussions. I might create a new one once we have a clearer picture of how this will get integrated into ITK (in-source vs npm dependency), [`dicomParser`](https://github.com/cornerstonejs/dicomParser/) vs [`daikon`](https://github.com/rii-mango/Daikon), etc...

Related discussions: [itk discourse](https://discourse.itk.org/t/extracting-header-information-from-dicom-image/2187/10), #250, #251 
cc: @thewtex, @floryst, @zachmullen, @pieper


examples/Dicom/src/parseDicomFiles.js
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  extractTags(dicomMetaData) {
+    const tags = this.constructor.tags
+    const primaryTag = this.constructor.primaryTag
+    tags.forEach((tag) => {
+      const value = dicomMetaData.string(DICOM_DICTIONARY[tag])
+      if (this[tag] === undefined) {
+        this[tag] = value
+      } else if (value != undefined) {
```
floryst: This should be fine since you're comparing with undefined, but FYI linters will likely want to see `!==` instead of `!=`.

zachmullen: I would go further and recommend that if you're attempting to compare with undefined, definitely use strict equality comparison, and if you only care about truthiness, be explicit and just do `if (value)`.

examples/Dicom/src/parseDicomFiles.js
```diff
@@ -182,13 +179,23 @@ const parseDicomFiles = async (fileList) => {
     patient.parseMetaData(dicomMetaData, file)
   }
 
+  // Set up promises
+  const parseFiles = [...fileList].map((file) => {
+    const promise = parseFile(file)
+    if (skipOnFailure) {
+      promise.catch((error) => {
+        failures.push({ file, error })
+      })
```
agirault: Even with the `catch` not throwing again (just storing the failure), I get an uncaught exception when calling [`Promise.all`](https://github.com/InsightSoftwareConsortium/itk-js/pull/252/commits/c41ac4cc26c5caf8b1eaf6d1c2ff2e4d1b124677#diff-62f35c82e16d848a11710ce525c9a4faR196)...

zachmullen: I think that's by design, does it work if you wrap up the Promise.all with try/catch in that case?

agirault: Thanks. Addressed in 1aa0c8807a0b68968a30a3d50df0535c5c7b37e8.

pieper:
Hi @agirault - 

It's a complex topic, as you know!  Especially the Slicer part, since many cases are handled.  Happy to talk if over if that would help.

Yes, for sure we should be treating all of this consistently and efficiently and I appreciate the effort you are putting into it.

We typically rely on the server to sort patient/study/series in the database (e.g. https://github.com/dcmjs-org/dicomweb-server uses CouchDB and Slicer uses SQLite via CTK https://github.com/commontk/CTK/tree/master/Libs/DICOM/Core.) 

Once you have the instances for a series, in dcmjs the mapping to volumes is here:

https://github.com/dcmjs-org/dcmjs/blob/master/src/normalizers.js

The idea is that a "normalized" dicom is a multiframe conversion, often referred to as a 'dataset' in the code.  This has all the image data in an array that is effectively exactly what ITK and VTK expect as pixel arrays.

Here's the code to map the dataset to a vtkImageData, which is simple as you can see.

https://github.com/dcmjs-org/dcmjs/blob/master/examples/vtkDisplay/index.html#L448-L475

I'm going to put on my `opinionated hat` here, and say that dcmjs does things the right way.  It is my favorite of all the implementations for a lot of reasons (that I'm happy to describe at length if prompted!).  In a nutshell dcmjs reflects a more comprehensive perspective on various quantitative imaging use cases.

-Steve

agirault:
Hi @pieper. Thanks a lot for sharing this.

> Once you have the instances for a series, in dcmjs the mapping to volumes is [here](https://github.com/dcmjs-org/dcmjs/blob/master/src/normalizers.js). The idea is that a "normalized" dicom is a multiframe conversion, often referred to as a 'dataset' in the code. This has all the image data in an array that is effectively exactly what ITK and VTK expect as pixel arrays. I'm going to put on my opinionated hat here, and say that dcmjs does things the right way. 

Indeed, it seems dcmjs does it right, and I don't want to reinvent the wheel so I'll hold off on continuing implementing this in this current example.

> We typically rely on the server to sort patient/study/series in the database (e.g. [dicomweb-server](https://github.com/dcmjs-org/dicomweb-server) uses CouchDB and Slicer uses SQLite via [CTK](https://github.com/commontk/CTK/tree/master/Libs/DICOM/Core).)

We need to be able to do this sorting client-side since in our use-case the user uploads DICOM files and selects a series that will be rendered on the client, and eventually sent over to the server afterward. My current proof-of-concept in this branch achieves that by leveraging `dicomParser`: do you see a way we could integrate that in the dcmjs workflow, to be able to combine the parsing of `dicomParser` and the `dcmjs` normalization of the selected series all on the client?

Also worth noting that the current implementation I have in this branch is about 6x (with reslice intercept and/or slope) to 17x faster (without reslice) than the ITK `readImageDICOMFileSeries`.

agirault:
FYI @finetjul

pieper:
@agirault sorry for the slow reply, I was out last week.

Regarding the browser-centric workflow I entirely agree.  I'd still stick with the database approach to organizing the series though, for consistency.  We haven't fleshed it out yet, but I would like to see the same dicomweb-server code applied via PouchDB so that the 'client' side of the browser can use DICOMweb API calls even though it is talking to a local facade.

I agree that doing this in native javascript is a better idea than using cross-compiled C++.  Not just for speed, but also because there are a lot of use cases for DICOM that still need work to represent well and there's no reason to limit ourselves to what has been implemented for imaging in the past.  We can address more general quantitative imaging tasks like segmentation and annotation more cleanly using the dcmjs approach.

Would you be interested in having a chat sometime to compare notes?  There are several projects that probably share some core needs and maybe we can make progress together.

agirault:
Updated to handle rescale intercept, rescale slope, and masking, and added a chrono to log timings and compare.

agirault:
Updated to speed up rescaling and parsing (less function calls, more classic loops), thanks @floryst for the suggestions.

agirault:
Updated to speed up rescaling and parsing (less function calls, more classic loops), thanks @floryst for the suggestions.

examples/Dicom/src/parseDicomFiles.js
```diff
@@ -0,0 +1,474 @@
...
+    const jDirCos = directionCosines.slice(3, 6)
+    const kDirCos = [
+      iDirCos[1] * jDirCos[2] - iDirCos[2] * jDirCos[1],
+      iDirCos[2] * jDirCos[0] - iDirCos[0] * jDirCos[2],
+      iDirCos[0] * jDirCos[1] - iDirCos[1] * jDirCos[0],
+    ]
+    const direction = {
+      rows: 3,
+      columns: 3,
+      data: [
```
WillCMcC, line 183: For me, parsing the same tags and using this code, direction.data ended up being: [...iDirCos, ...jDirCos, ...kDirCos]. Just a heads up!

thewtex:
For compatibility with DICOM, AWS, etc, the metadata could be structured similar to the ImageSet metadata described here:

https://github.com/aws-samples/amazon-healthlake-imaging-samples/blob/88666ee00a0830b3a7cab403267887b8ff6de662/pixel-data-verification/index.js#L43-L49

PR #282, opened 2019-11-15
Modify itk-js-cli.js to call git bash on windows

HastingsGreer:
Currently when "npx itk-js build ." is called on windows, itk-js-cli.js attempts to call web-build/itk-js-build-env using cmd. This commit modifies its behavior to use git bash instead. It also adds some logging so that if spawnSync fails, a helpful error message is printed instead of failing silently.


thewtex:
Published in itk.js 10.2.1 :tada: :taco:

PR #297, opened 2020-03-19
doc(Image): Add note about indexing convention

### LINKED ISSUE #296: Unpacking Image.data
When reading a DICOM using one of [these methods](https://insightsoftwareconsortium.github.io/itk-js/api/browser_io.html) we get `itk/Image` object. But when trying to unpack the `Image.data` there is one problem - the `Image.size` stores indexes in ITK’s format (x,y,z) but the `Image.data` looks like a flatten multidimensional array that had indexes in order (z,x,y). So, in order to unpack it, you have to be aware of the history of these transformations and read it like this:

`var tensor = tf.tensor(image.data, [image.size[2], image.size[0], image.size[1]]);`

As you can see it's a bit confusing and requires some debugging, tweaking and looking into data to figure this out. Maybe you can add a property like `dataShape`, so reading will look like `tf.tensor(image.data, image.dataShape);`? Or reshape the array to `image.size` format before flattening.

But maybe I just missed something and there is another way to do it.

thewtex:
Closes #296

### DIFF
--- a/doc/content/api/Image.md
+++ b/doc/content/api/Image.md
@@ -1,7 +1,7 @@
 title: Image
 ---
 
-An `itk/Image` is the N-dimensional image data structure for *itk.js*. To visualize an `itk/Image` with *vtk.js*, use `convertItkToVtkImage` from [`vtk.js/Sources/Common/DataModel/ITKHelper`](https://kitware.github.io/vtk-js/api/Common_DataModel_ITKHelper.html). `itk/Image` is a JavaScript object with the following properties:
+An `itk/Image` is the N-dimensional image data structure for *itk.js*. `itk/Image` is a JavaScript object with the following properties:
 
 **imageType**: The [ImageType](./ImageType.html) for this image.
 **name**: An optional name, a `String`, that describes this image.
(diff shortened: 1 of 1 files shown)


thewtex:
@HastingsGreer please take a look

HastingsGreer:
Lgtm!

HastingsGreer:
On further inspection, I do have a question: the issue states that Serhiy solved his problem by reordering the size [2, 0, 1] but the new documentation suggests reversing it, which is [2, 1, 0]. Does it work either way?

thewtex:
@HastingsGreer, yes -- @Serhiy-Shekhovtsov in #296 , is the ordering:

`(z,x,y)` correct?

Serhiy-Shekhovtsov:
@HastingsGreer , @thewtex sorry for confusing you guys. I don't know where I got the idea it goes x,y vs y,x. It looks like the ordering is a simple reverse.
There is another problem, however - slices go in opposite order when using itk-js vs SimpleITK(python). The first slice in JS is the last slice in Python. You can read more about it [here](https://discourse.itk.org/t/imageseries-reader-reads-dicom-series-in-reverse-order-by-default/2491). Also here are some screenshots:
## JS
image
## Python
image

thewtex:
@Serhiy-Shekhovtsov NumPy also uses [z, y, x] indexing. This is explained here:

  https://scikit-image.org/docs/dev/user_guide/numpy_images.html#numpy-indexing

The [ITK Python wrapping is adding xarray support](https://discourse.itk.org/t/itk-5-1-release-candidate-2-available-for-testing/2803/4), in which case you can explicitly index the dimension by name, i.e. `'x'`, which avoids confusion and errors related to this issue.

Serhiy-Shekhovtsov:
Thanks @thewtex , now it's clear!

PR #307, opened 2020-03-31
fix(IO)

dzenanz:


thewtex:
Thanks for the patch. As indicated by CI, this is intentional since JavaScript/WebAssembly/Emscripten does not have full 64-bit integer support.

PR #316, opened 2020-04-09
feat(readImageDICOMFileSeries): Use GDCM instead of DCMTK

thewtex:
More transfer syntax support.

### DIFF
--- a/test/Browser/DICOMSeriesTest.js
+++ b/test/Browser/DICOMSeriesTest.js
@@ -24,7 +24,7 @@ test('Test reading DICOM file series', t => {
     })
     .then(function ({ image, webWorkerPool }) {
       t.is(image.imageType.dimension, 3, 'dimension')
-      t.is(image.imageType.componentType, IntTypes.UInt8, 'componentType')
+      t.is(image.imageType.componentType, IntTypes.Int16, 'componentType')
       t.is(image.imageType.pixelType, PixelTypes.Scalar, 'pixelType')
       t.is(image.imageType.components, 1, 'components')
       t.is(image.origin[0], -17.3551, 'origin[0]')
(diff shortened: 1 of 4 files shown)


test/Browser/DICOMSeriesTest.js
```diff
@@ -24,7 +24,7 @@ test('Test reading DICOM file series', t => {
     })
     .then(function ({ image, webWorkerPool }) {
       t.is(image.imageType.dimension, 3, 'dimension')
-      t.is(image.imageType.componentType, IntTypes.UInt8, 'componentType')
+      t.is(image.imageType.componentType, IntTypes.Int16, 'componentType')
```
agirault, line 27: Why would those change? The pixel type should be defined by [PixelRepresentation](https://dicom.innolitics.com/ciods/ct-image/image-pixel/00280103) and [BitsAllocated](https://dicom.innolitics.com/ciods/ct-image/image-pixel/00280100).

test/DICOMSeriesTest.js
```diff
@@ -70,7 +70,8 @@ test('Test reading a DICOM file with the synchronous API', t => {
 test('Test reading a DICOM file with the synchronous API, assume sorted', t => {
   const image = readImageLocalDICOMFileSeriesSync(testFiles.sort(), true)
   t.is(image.imageType.dimension, 3, 'dimension')
-  t.is(image.imageType.componentType, IntTypes.UInt8, 'componentType')
+  t.is(image.imageType.componentType, IntTypes.Int16, 'componentType')
+  t.is(image.imageType.pixelType, PixelTypes.Scalar, 'pixelType')
```
agirault, line 74: duplicated line `t.is(image.imageType.pixelType, PixelTypes.Scalar, 'pixelType')`

test/Browser/DICOMSeriesTest.js
thewtex: Likely related to how rescale slope / intercept is applied.

test/DICOMSeriesTest.js
thewtex: Good catch and thanks for the review! Removed here: https://github.com/InsightSoftwareConsortium/itk-js/pull/322

Issue #248, opened 2019-06-26
Add streaming support to itk::JSONImageIO


thewtex:
Add streaming support to [itk::JSONImageIO](https://github.com/InsightSoftwareConsortium/itk-js/blob/master/include/itkJSONImageIO.h) to help support processing very large images in wasm.

The [itk::StreamingImageIOBase](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/ImageBase/include/itkImageIOBase.h) streaming should be implemented [as described here](https://www.insight-journal.org/browse/publication/729).

thewtex:
Addressed in #334

Issue #213, opened 2019-03-18
Multithreading support


thewtex:
Enabling building modules with multi-threading support and using these modules on clients that support it.

PR #334, opened 2020-04-29
feat(WorkerPool): Support reporting progress

thewtex:
Also:

- Adds streaming support to itk::JSONImageIO
- Adds the copyImage function
- WebWorkerPoolTest is a good example basis for multi-threading an ITK processing pipeline.


thewtex:
Closes #248 #213

PR #366, opened 2020-08-28
Disable threads for openjp2

floryst:
This should fix https://github.com/Kitware/paraview-glance/issues/367.

Issue: openjp2 expects threads to be on by default, so it hangs while waiting for threads to initialize (which they never will, presently). This affects loading jpeg2000 images.

The `OPJ_USE_THREAD:BOOL=OFF` flag didn't work for me initially, so I change the openjp2 CMake file to turn off the threads option. Is there a better way of ensuring the option gets applied to gdcm's openjp2 lib? I can try again with just the flag (and without the sed hackery).

### DIFF
--- a/src/Docker/itk-js-base/Dockerfile
+++ b/src/Docker/itk-js-base/Dockerfile
@@ -11,6 +11,8 @@ RUN git clone https://github.com/InsightSoftwareConsortium/ITK.git && \
   cd ITK && \
   git checkout ${ITK_GIT_TAG} && \
   cd ../ && \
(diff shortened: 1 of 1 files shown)


thewtex (approved):
@floryst :100: :1st_place_medal: thanks!

agirault:
> The OPJ_USE_THREAD:BOOL=OFF flag didn't work for me initially, so I change the openjp2 CMake file to turn off the threads option. Is there a better way of ensuring the option gets applied to gdcm's openjp2 lib? I can try again with just the flag (and without the sed hackery).

That indeed seems hacky. @thewtex any opinion on this?

Issue #369, opened 2020-09-03
TypeError: imageIO.CanWriteFile is not a function when running itk-convert ./input.nrrd ./output.vti


mohghaderi:
itk-convert ./input.nrrd ./output.vti
in the example won't work and gives this error:

TypeError: imageIO.CanWriteFile is not a function
    at writeImageLocalFileSync (C:\TEMP\Kitware\itk-js\examples\Node.js\node_modules\itk\writeImageLocalFileSync.js:52:19)
    at writeLocalFileSync (C:\TEMP\Kitware\itk-js\examples\Node.js\node_modules\itk\writeLocalFileSync.js:35:5)
    at Object.<anonymous> (C:\TEMP\Kitware\itk-js\examples\Node.js\src\itk-convert.js:24:3)
    at Module._compile (internal/modules/cjs/loader.js:1156:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1176:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
    at internal/main/run_main_module.js:18:47

The problem is  writeImageLocalFileSync  iterate in all modules and can't find CanWriteFile function in itk\ImageIOs\itkDCMTKImageIOJSBinding module.

thewtex:
Update for the example incorporating #372 is added in #375

PR #370, opened 2020-09-03
#369 Fix CanWriteFile is not a function

mohghaderi:
ImageIO modules don't have `CanWriteFile` function defined, it causes the module to throw an errors
this fix skip modules that don't have this function


thewtex:
@mohghaderi thanks for reporting the issue and contributing the PR!!

It turns out that all ImageIO's should be exposing `CanWriteFile` and a few are missing the binding -- they are added in #372

PR #395, opened 2021-02-09
Support cancel pending tasks which have not yet run

scottwittenburg:
This PR changes the behavior of the `WorkerPool.runTasks` method so instead of returning a promise, it returns an object containing the original promise along with an id that can be used to cancel pending tasks before they have been executed by a worker (cancelation is achieved via a new `cancel(runId)` method added to the `WorkerPool` api).

### DIFF
--- a/src/WorkerPool.js
+++ b/src/WorkerPool.js
@@ -19,9 +19,13 @@ class WorkerPool {
    * Run the tasks specified by the arguments in the taskArgsArray that will
    * be passed to the pool fcn.
    *
-   * An optional progressCallback will be cassed with the number of complete
+   * An optional progressCallback will be called with the number of complete
    * tasks and the total number of tasks as arguments every time a task has
    * completed.
+   *
+   * Returns an object containing a promise ('promise') to communicate results
+   * as well as an id ('runId') which can be used to cancel any remaining pending
+   * tasks before they complete.
    */
   runTasks (taskArgsArray, progressCallback) {
     const info = {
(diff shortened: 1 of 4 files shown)


src/WorkerPool.js
```diff
@@ -59,14 +67,26 @@ class WorkerPool {
...
+    }
+  }
+
   // todo: change to #addTask(resultIndex, taskArgs) { after private methods
   // proposal accepted and supported by default in Babel.
   addTask (infoIndex, resultIndex, taskArgs) {
     const info = this.runInfo[infoIndex]
+
+    if (info && info.canceled) {
+      info.reject('Remaining tasks canceled')
```
thewtex, line 83: *cancelled*

thewtex: We could also add cancel checks around line 74 and line 115.

src/WorkerPool.js
```diff
@@ -30,23 +34,27 @@ class WorkerPool {
       addingTasks: false,
       postponed: false,
       runningWorkers: 0,
-      progressCallback: progressCallback
+      progressCallback: progressCallback,
+      canceled: false
```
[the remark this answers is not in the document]
scottwittenburg: It's fine with me if it's your preference, but technically it's not misspelled: https://www.dictionary.com/e/canceled-vs-cancelled/

Just saying 😉

src/WorkerPool.js
scottwittenburg: > We could also add cancel checks around line 74 and line 115.

Yeah that makes sense, I'll add them there as well.

src/WorkerPool.js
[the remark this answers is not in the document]
thewtex: Ah, let's use the American English version, then ;-)

scottwittenburg:
@thewtex So this seems to be working really well for my application now.  I know we discussed *not* nullifying the slots in `runInfo`, but I'm guarding access to those elements everywhere now (I think), and I'm not getting those hard-to-diagnose errors anymore.  Also, the memory footprint is no longer monotonically increasing after starting lots of task lists and canceling many of them.  

Anyway, in theory at least, nullifying the slots in `runInfo` seems like the right thing to do to free up as much memory as possible when a  task ends for whatever reason.  As an aside, it also seemed to help with memory consumption to nullify all the heavy fields in each `runInfo` element before nullifying the slot itself.

Let me know what you think.

thewtex:
@scottwittenburg awesome!!! :rocket: 

Could you please collect the updates in the PR?

scottwittenburg:
> Could you please collect the updates in the PR?

Sorry, what do you mean here?  I'm sure I should know, but for some reason I'm drawing a blank.

thewtex (approved):

scottwittenburg:
Thanks @thewtex!  🎉 🎊 🚀

Issue #394, opened 2021-02-08
Add `metadata` property to itk.Image and populate it from file metadata


thewtex:
E.g. DICOM tags

agirault:
cc: @floryst

floryst:
Some thoughts:
- DICOM tags are associated with DICOM objects. When the objects are slices, then it could be the case that certain tags (e.g. RescaleSlope and RescaleIntercept) vary between each slice/object. We would want to allow requesting the DICOM tags for a particular object.
- There can be a lot of tags, and I don't know if we want to incur the performance penalty (if significant; needs testing) of reading all tags up-front. We could make such a `metadata` property into a `getDicomTag` method, but we still need to choose what DICOM object to read from.
- If we read a single slice, then `metadata` makes more sense than if we constructed a volume.

One idea is to have a function like so: `readDICOMTags(dicomFile, [...tagNames])`. This is likely the smallest useful primitive for reading DICOM tags. The problem with this approach is that there will be double-parsing of the dicom header when the user also wants to render said DICOM as part of a volume. I suppose that shouldn't be too bad considering that ITK's `DicomIO.ReadInformation` (I think that's what it's called) only reads the DICOM header.

If we want a property off of itk.Image, maybe it can be something like `image.getDICOMTag(slice#, [...tagNames])`. If the itk.Image is a volume, then this assumes we have a mapping from volume (Z) slice to dicom object, unless we extract all the tags per slice during volume construction. If the itk.Image is a slice, then all we need is `image.getDICOMTags(...tagNames)`.

The classic use-case: User wants to scroll through a DICOM stack and have the W/L automatically set from the RescaleSlope and Intercept of the current slice.
- If we are slicing a 3D volume, how will we map from a slice # to the correct set of DICOM tags?
- Is it more worthwhile to get the correct ordering of slices, then read each slice into its own image w/ tags? AFAIK DICOM viewers take this approach, I think.

agirault:
I agree that important issues to take into account are:
1. that some tags vary across single images/scans (ImagePosition, TriggerTime...)
2. that we might not want to keep the whole metadata in memory
3. another one to take into account: how do we handle nested datasets/metadata ([sequences](http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_7.5.html))

> image.getDICOMTag(slice#, [...tagNames])

[...]

PR #406, opened 2021-03-26
Read dicom tags

floryst:
This adds an API for reading DICOM tags from DICOM object files. Additionally, appropriate character set conversion is performed, decoding encoded strings into UTF-8.

Relevant issue: #394 

Todo:
- [x] add async version
- [x] add browser version
- [x] add charset decoder
- [x] Add docs
- [x] Make API and docs consistent

### DIFF
--- a/src/readDICOMTags.js
+++ b/src/readDICOMTags.js
@@ -1,0 +1,32 @@
+import createWebworkerPromise from './createWebworkerPromise'
+import PromiseFileReader from 'promise-file-reader'
+
+import config from './itkConfig'
+
+const readDICOMTags = async (webWorker, file, tags = null) => {
+  let worker = webWorker
+  const { webworkerPromise, worker: usedWorker } = await createWebworkerPromise(
+    'ImageIO',
+    worker
+  )
+  worker = usedWorker
+  const arrayBuffer = await PromiseFileReader.readAsArrayBuffer(file)
+  try {
+    const tagValues = await webworkerPromise.postMessage(
+     {
+        operation: 'readDICOMTags',
+        name: file.name,
+        type: file.type,
+        data: arrayBuffer,
+        tags: tags,
+        config: config
+      },
+      [arrayBuffer]
+    )
+    return { tags: tagValues, webWorker: worker }
+  } catch (error) {
+    throw Error(error)
+  }
+}
+
+export default readDICOMTags
--- a/doc/content/api/browser_io.md
+++ b/doc/content/api/browser_io.md
@@ -69,6 +69,19 @@ If the files are known to be from a single, sorted series, the last argument can
 
 The used `webWorkerPool` is returned to enable resource cleanup by calling `.terminateWorkers()`.
 
+## readDICOMTags(webWorker, file, tags=[]) -> { tags, webWorker }
+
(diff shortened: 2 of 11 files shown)


floryst:
FYI @agirault

agirault:
Can you add usage documentation @floryst ? or an example?

src/readDICOMTags.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  const tagResults = await webworkerPromise.postMessage(message, transferables)
+  return { tags: tagResults, webWorker: worker }
+}
+const numberOfWorkers = navigator.hardwareConcurrency ? navigator.hardwareConcurrency : 4
+const workerPool = new WorkerPool(numberOfWorkers, workerFunction)
+
+const readDICOMTags = async (
+  fileList,
+  tags = null
```
floryst, line 39: These parameters may change, pending API discussions.
- fileList: probably going to become a single file rather than a list of files
- tags: if we want a null/undefined `tags` arg to mean "give me all tags", then this will stay. Right now there is no logic for getting all tags.

thewtex: +1 for the single file. Specification of the tags-per-file is simpler this way, too. +1 for the current `null` returns all tags API.

thewtex (changes requested):
@floryst awesome!! :rocket: :man_singer: 

> Additionally, appropriate character set conversion is performed, decoding encoded strings into UTF-8.

This is fantastic! Can we add a test to ensure its functionality?

> I'm not sure why I can't request reviewers, so here: @thewtex @agirault.

@floryst after accepting the invitation to the repository (just sent), you should be able to request reviewers.

 >  I think the API would make more sense if we simply read the tags from a single file instead

+1

> Would it make sense to support returning all tags, rather than specifying which ones the user wants? If so, I think setting tags=null|undefined can mean "return all tags".

+1

> I haven't tested private tags. What would the expected behavior be for private tags, since we don't necessarily know the tag's VR? (This is partially a GDCM question.)

In the future, if this is possible and needed, we could add a third argument, `privateTags=false`.

> Code style: is there an automatic c++ style enforcer for itk-js? My c++ code currently has a mix of the itk-js c++ coding style and the clang c++ codingstyle and I wanted to ask before making manual changes to match the itk-js c++ coding style.

Thanks for asking! The code is looking very clean, btw. We should use the ITK clang-format style. To apply it to the current code, download the [clang-format-8](https://data.kitware.com/#collection/57b5c9e58d777f126827f5a1/folder/5d635cbcd35580e6dcbd80b6) executable, and run it on the source code files with the *ITK/.clang-format* configuration file.

We gave a pre-commit hooks setup for ITK that runs clang-format -- we could add this to itk.js sometime.

> libiconv: if the docker container installed the requisite compilation headers for iconv (which might be provided by libc6 rather than a dedicated package), then I wouldn't need to add iconv as an external project. As it stands, the ExternalProject is the only way to use iconv, which is why I include it.

This is awesome. While we could add the package to the docker image, I think the ExternalProject is better because we can change the base image Linux distribution and Linux distribution version, and not have to worry about losing the required headers or them moving around. We can also independently update the libiconv version.

thewtex (approved):
@floryst beautiful!

PR #408, opened 2021-04-20
style(readDICOMTags): Update to make standard lint check pass

thewtex:
Applied with npx standard --fix src/readDICOMTags.js.


thewtex:
@floryst with this fix, all tests pass locally for me 💯

PR #467, opened 2022-02-01
refactor(ReadDICOMTags): Refactor as pipeline

thewtex:
Efficiency, use outside Emscripten, avoid EmBind memory issues.


thewtex:
@floryst @sankhesh @jjomier

PR #471, opened 2022-02-04
fix(readDICOMTags): import InterfaceTypes directly

### LINKED ISSUE #470: Remove dependency from node built-ins
webpack 5 doesn't polyfill node built-ins anymore (https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-nodejs-polyfills-removed)

This library seems to depend on many modules that require these polyfills, would you consider to replace them with browser-friendly versions?

floryst:
Importing from '../index.js' pulls in node imports, which is not
desirable for web-based distribution. Instead of importing from index, import directly from InterfaceTypes.

Fixes #470.

### DIFF
--- a/src/io/readDICOMTags.ts
+++ b/src/io/readDICOMTags.ts
@@ -6,7 +6,7 @@ import TextStream from '../core/TextStream.js'
 import config from '../itkConfig.js'
 
 import ReadDICOMTagsResult from './ReadDICOMTagsResult.js'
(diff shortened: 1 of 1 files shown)


thewtex (approved):
Thanks for the fix, Forrest!!

FezVrasta:
Could we get a new npm release with this change please?

FezVrasta:
`itk-wasm@1.0.0-a.12` includes these changes

PR #479, opened 2022-02-08
fix: remove console.log from pipeline worker

FezVrasta:
this unnecessary adds logging to the console and slows down the process when the dev tools are open


thewtex (approved):
@FezVrasta thank you! :pray:

PR #490, opened 2022-02-24
.iwi.cbor.zstd Image IO support

thewtex:
- build(CMake): Build and link zstd lib
- feat(Compress): Add CompressStringify
- feat(Compress): Add ParseStringDecompress


thewtex:
@WillCMcC @sankhesh this is now available in v1.0.0-a14, interested to know how well it works for you.

PR #491, opened 2022-02-25
fix(WorkerPool): Address memory leak in Chrome from reject/resolve

thewtex:
On Chrome repeated calls to the worker pool can accumulate leaks due to
reject / resolve holding a reference to results even after the
associated Promise has resolved.

Co-authored-by: Forrest Li <forrest.li@kitware.com>
Co-authored-by: Will McCambley <willmccambley@gmail.com>
Co-authored-by: Sankhesh Jhaveri <sankhesh.jhaveri@kitware.com>

### DIFF
--- a/src/core/WorkerPool.ts
+++ b/src/core/WorkerPool.ts
@@ -1,3 +1,5 @@
+/* eslint-disable  @typescript-eslint/no-non-null-assertion */
+
 import WorkerPoolFunction from './WorkerPoolFunction.js'
 import WorkerPoolProgressCallback from './WorkerPoolProgressCallback.js'
 import WorkerPoolRunTasksResult from './WorkerPoolRunTasksResult.js'
@@ -104,9 +106,8 @@ class WorkerPool {
     const info = this.runInfo[infoIndex]
 
     if (info?.canceled === true) {
+      info.reject!('Remaining tasks canceled')
       this.clearTask(info.index)
-      // @ts-expect-error: error TS2722: Cannot invoke an object which is
-      info?.reject('Remaining tasks canceled')
       return
     }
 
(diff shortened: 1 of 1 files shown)


thewtex:
@floryst @sankhesh @WillCMcC please take a look

src/core/WorkerPool.ts
```diff
@@ -104,9 +104,9 @@ class WorkerPool {
     const info = this.runInfo[infoIndex]
 
     if (info?.canceled === true) {
-      this.clearTask(info.index)
       // @ts-expect-error: error TS2722: Cannot invoke an object which is
       info?.reject('Remaining tasks canceled')
+      this.clearTask(info.index)
```
sankhesh, line 109: Do you need `info?.index`?

src/core/WorkerPool.ts
```diff
@@ -104,9 +104,9 @@ class WorkerPool {
     const info = this.runInfo[infoIndex]
 
     if (info?.canceled === true) {
-      this.clearTask(info.index)
       // @ts-expect-error: error TS2722: Cannot invoke an object which is
```
sankhesh, line 107: Missing the tail end of the comment

```
 // @ts-expect-error: error TS2722: Cannot invoke an object which is possibly 'undefined'.
```

WillCMcC:
Ran these changes on our test program that runs readImageDICOMArrayBufferSeries in a loop-- the issue does appear to be fixed 👍

FezVrasta:
Would it make sense to add a comment on those lines to explain why the order is important? I feel like it could get easily accidentally reverted during reactors otherwise

src/core/WorkerPool.ts
thewtex: The `?` on `info?.reject` was not needed :-) -- removed now.

thewtex:
> 
Would it make sense to add a comment on those lines to explain why the order is important? I feel like it could get easily accidentally reverted during reactors otherwise

:+1:  added

PR #494, opened 2022-02-27
feat(package.json): Add export maps

thewtex:
See:

- https://nodejs.org/api/packages.html#exports-sugar
- https://github.com/parcel-bundler/parcel/issues/4155

Suggested-by: Federico Zivolo <federico.zivolo@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,18 @@
   "version": "1.0.0-a.14",
   "description": "High-performance spatial analysis in a web browser, Node.js, and reproducible execution across programming languages and hardware architectures.",
   "main": "./dist/index.js",
+  "browser": {
+    "./index.js": "./dist/browser/index.js"
+  },
+  "exports": {
+    ".": {
+      "browser": "./dist/browser/index.js",
+      "import": "./dist/node/index.js",
+      "umd": "./dist/umd/itk-wasm.js",
+      "package.json": "./package.json",
+      "./index.d.ts": "./dist/index.d.ts"
+    }
+  },
   "type": "module",
   "types": "./dist/index.d.ts",
   "directories": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -3,6 +3,13 @@
   "version": "1.0.0-a.14",
   "description": "High-performance spatial analysis in a web browser, Node.js, and reproducible execution across programming languages and hardware architectures.",
   "main": "./dist/index.js",
+  "exports": {
+    ".": {
```
FezVrasta, line 7: This will prevent to import other modules individually. If I'm not wrong they should all be listed here

thewtex: Yes, my understanding also is that this will prevent the import of other modules individually. However, all exposed modules should be available through the index. Adding all modules would add a maintainability headache. From the other examples I looked at:

- https://unpkg.com/browse/preact@10.6.6/package.json
- https://unpkg.com/browse/nanoid@3.3.1/package.json

only the package index's are exposed. Is there really a need to list all the modules?

FezVrasta: I'm from mobile so I can't look for the exact name, but in our case we need the worker pool utility that AFAIK is not exposed from the index module

thewtex: Ah, ok. The worker pool should be exposed as

```
import { WorkerPool } from 'itk-wasm'
```

per:

https://github.com/InsightSoftwareConsortium/itk-wasm/blob/6539760e0d0a70eba265c6dd2b661331cf15a4ef/src/core/index.ts#L35-L38

https://github.com/InsightSoftwareConsortium/itk-wasm/blob/6539760e0d0a70eba265c6dd2b661331cf15a4ef/src/index.ts#L3

If not, let's fix it and make sure it meets your needs.

FezVrasta: I'll double check tomorrow, thanks!

FezVrasta: It works, thanks 🙇

PR #497, opened 2022-03-01
build(io-packages): Remove dependency on itk-wasm

thewtex:
These are optional dependencies for itk-wasm. Avoid pulling down
node_modules/itk-image-io/node_modules/itk-wasm/*


thewtex:
@FezVrasta @sankhesh please take a look.

sankhesh (approved):

PR #512, opened 2022-03-08
fix(WASMMeshToMeshFilter): Assign all points

thewtex:


include/itkWASMPolyDataToPolyDataFilter.hxx
```diff
@@ -189,8 +189,8 @@ WASMPolyDataToPolyDataFilter<TPolyData>
     const rapidjson::Value & pointsJson = document["points"];
     const std::string pointsString( pointsJson.GetString() );
     CoordRepType * pointsPtr = reinterpret_cast< CoordRepType * >( std::atol(pointsString.substr(35).c_str()) );
-    polyData->GetPoints()->resize(numberOfPoints);
-    polyData->GetPoints()->assign(pointsPtr, pointsPtr + numberOfPoints);
+    polyData->GetPoints()->resize(numberOfPoints * 3);
+    polyData->GetPoints()->assign(pointsPtr, pointsPtr + numberOfPoints * 3);
```
N-Dekker, line 193: @thewtex Do you mean something like:

    CoordRepType * coordPtr = reinterpret_cast<CoordRepType *>(std::atoll(pointsString.substr(35).c_str()));

    for (PointType & point : *(polyData->GetPoints()))
    {
      for (auto & coord : point)
      {
        coord = *coordPtr;
        ++coordPtr;
      }
    }

I'm sorry, but I think it cannot be written as a single `assign` call, because the right side consists of individual coordinate values, while the left side consists of points.

include/itkWASMMeshToMeshFilter.hxx
```diff
@@ -209,8 +209,8 @@ WASMMeshToMeshFilter<TMesh>
   const rapidjson::Value & pointsJson = document["points"];
   const std::string pointsString( pointsJson.GetString() );
   CoordRepType * pointsPtr = reinterpret_cast< CoordRepType * >( std::atol(pointsString.substr(35).c_str()) );
-  mesh->GetPoints()->resize(numberOfPoints);
-  mesh->GetPoints()->assign(pointsPtr, pointsPtr + numberOfPoints);
+  mesh->GetPoints()->resize(numberOfPoints * dimension);
+  mesh->GetPoints()->assign(pointsPtr, pointsPtr + numberOfPoints * dimension);
```
Leengit, line 213: I didn't check, so apologies if this is moot ... if `resize` and `assign` support `numberOfPoints` and `PointType *` rather than `numberOfCoordinates` and `CoordRepType *` then
```cpp
  using PointType = typename MeshType::PointType;
  const auto * const pointsPtr = reinterpret_cast< PointType * >( std::atol(pointsString.substr(35).c_str()) );
  mesh->GetPoints()->resize(numberOfPoints);
  mesh->GetPoints()->assign(pointsPtr, pointsPtr + numberOfPoints);
```

thewtex: Yes, that's!

include/itkWASMPolyDataToPolyDataFilter.hxx
thewtex: The correct changes is the one suggested by @Leengit below.

include/itkWASMMeshToMeshFilter.hxx
```diff
@@ -205,10 +205,10 @@ WASMMeshToMeshFilter<TMesh>
     throw std::runtime_error("Unexpected cell pixel type");
   }
 
-  using CoordRepType = typename MeshType::CoordRepType;
+  using PointType = typename MeshType::PointType;
   const rapidjson::Value & pointsJson = document["points"];
   const std::string pointsString( pointsJson.GetString() );
-  CoordRepType * pointsPtr = reinterpret_cast< CoordRepType * >( std::atol(pointsString.substr(35).c_str()) );
+  const auto * pointsPtr = reinterpret_cast< PointType * >( std::atol(pointsString.substr(35).c_str()) );
```
N-Dekker, line 211: I still wonder if it's OK to represent a pointer by a signed long integer, as retrieved by `std::atol`. On Windows, a `long`is only 32 bits, you know. Maybe better use `std::strtoull` instead of  `std::atol`?  `std::strtoull` converts to 64-bit unsigned long long, which should be sufficient for a `PointType *` pointer on any platform. Anyway, just my two cents!

thewtex: Good thoughts, @N-Dekker . WASM is current 32-bit, but this could be improved for the future. I created #515 to track.

PR #529, opened 2022-03-11
improvement(DICOM IO): Add default restrictions to DICOM reader

sankhesh:
Even though UseSeriesDetails was set to true, the default restrictions were missing.
These restrictions are used to sub-refine the DICOM file set into different series.


sankhesh:
@thewtex PTAL

thewtex (approved):
@sankhesh :+1: thank you!

Issue #524, opened 2022-03-10
Performance issue for large itk images (Reader and Writer)


Guzzler:
#include<chrono>

using namespace std::chrono;
#include "itkImage.h"

#include "itkImageFileReader.h"
#include "itkImageFileWriter.h"

int main(int argc, char *argv[])
{
  auto start = high_resolution_clock::now();
  if (argc < 3)
  {
    std::cerr << "Missing Parameters " << std::endl;
    std::cerr << "Usage: " << argv[0];
    std::cerr
        << "inputImage  outputImage"
        << std::endl;
    return EXIT_FAILURE;
  }

  using InternalPixelType = float;
  constexpr unsigned int Dimension = 3;
  using InternalImageType = itk::Image<InternalPixelType, Dimension>;

  using OutputPixelType = float;
  using OutputImageType = itk::Image<OutputPixelType, Dimension>;

  using ReaderType = itk::ImageFileReader<InternalImageType>;
  using WriterType = itk::ImageFileWriter<OutputImageType>;

  auto reader = ReaderType::New();
  auto writer = WriterType::New();

  reader->SetFileName(argv[1]);
  writer->SetFileName(argv[2]);

  writer->SetInput(reader->GetOutput());
  
  try
  {
    writer->Update();
  }
  catch (const itk::ExceptionObject &excep)
  {
    std::cerr << "Exception caught !" << std::endl;
    std::cerr << excep << std::endl;
  }
  auto stop = std::chrono::high_resolution_clock::now();

  auto duration = duration_cast<microseconds>(stop - start);
 
  std::cout << "Time taken by function: "
        << duration.count() << " microseconds" << std::endl;

  return EXIT_SUCCESS;
}`

Guzzler:
Here is a basic example of the cxx pipeline that i built using itk-js version 14.1.1.
A simple ITK image with an arraybuffer of length (512x512x200) is taking 6 seconds to execute while using runPipelineBrowser. Is there anything that is being done wrong? I am reading and writing to a file called 'input.json' and 'output.json' respectively and passing both as ITKImage objects. Would appreciate some help on why it takes that long to simply read and write the files and how this can be optimized.

thewtex:
@Guzzler please try itk-wasm :-)

Guzzler:
@thewtex I had earlier attempted to build using itk-wasm but it was failing so i defaulted to itk.js 14.1.1. will the speed up with itk wasm be significant?

[...]

PR #533, opened 2022-03-16
pipeline full path

thewtex:
- refactor(createWebWorkerPromise): Remove name argument
- build(Convert{Image,Mesh}): Update factory register manager CMake variables
- refactor(createWebWorkerPromise): Use pipelineWorkerUrl with full path to worker


thewtex:
@Guzzler based on our discussion in #524 , this will make it easier to use a non-minified worker for debugging.

Guzzler:
This looks great @thewtex ! Will definitely use this as I build my pipelines.

PR #559, opened 2022-04-12
Release config

floryst:
This MR adds the semantic-release configuration to itk-wasm. The configuration details are provided below.
- Merges and commits to the `release` branch will publish to the default distribution channel, i.e. `npm install itk-wasm`. The tag format is `itk-wasm-v{version}`.
  - After this MR, adding a commit to the `release` branch will release itk-wasm v1.0.0.
- Merges and commits to the `master` branch will publish to the beta distribution channel, i.e. `npm install itk-wasm@beta`. The tag format is `itk-wasm-v{version}.b{betaVersion}`.
  - After this MR, adding a major/minor/patch commit to the `master` branch will release itk-wasm v1.0.0.b9, since the current beta version is b8.
- Semantic release will only run on the `master`, `release`, and any designated maintenance branches.
- There is no concept of an alpha release on master. If alpha releases are desired in the future, then a new branch will be needed.

### DIFF
--- a/.github/workflows/javascript-typescript.yml
+++ b/.github/workflows/javascript-typescript.yml
@@ -66,6 +66,15 @@ jobs:
 
         cd -
 
+    - name: Publish task
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      run: |
+        git config --global user.name "GitHub Actions"
+        git config --global user.email "itk+community@discourse.itk.org"
+        npx semantic-release
+
     - name: Setup ipfs
       uses: ibnesayeed/setup-ipfs@master
       with:
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,7 @@
         "readable-stream": "^3.6.0",
         "rollup": "^2.58.0",
         "rollup-plugin-terser": "^7.0.2",
+        "semantic-release": "^19.0.2",
         "standard": "^16.0.4",
         "stream-browserify": "^3.0.0",
         "tap-spec": "^5.0.0",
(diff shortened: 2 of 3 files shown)


.github/workflows/javascript-typescript.yml
```diff
@@ -66,6 +66,15 @@ jobs:
 
         cd -
 
+    - name: Publish task
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
floryst, line 72: @thewtex you will need to add an NPM_TOKEN to the repo secrets in order to publish.

Do you want to also publish to github? If not, we can disable the github plugin.

.github/workflows/javascript-typescript.yml
```diff
@@ -66,6 +66,15 @@ jobs:
...
+      run: |
+        git config --global user.name "Github Actions"
+        git config --global user.email "matt.mccormick@kitware.com"
```
floryst, line 75: Is there an itk robot account?

floryst:
Another note: due to how the beta channel works, adding more commits to `master` will result in `npm install itk-wasm` not installing the latest updates. That will have to be through `npm install itk-wasm@beta`.

.github/workflows/javascript-typescript.yml
thewtex: :1st_place_medal: :rocket: :man_singer: 

NPM_TOKEN added.

Sure, let's publish to GitHub, too.

.github/workflows/javascript-typescript.yml
thewtex: `Github` -> `GitHub`

We can use `itk+community@discourse.itk.org` for the email address.

thewtex (changes requested):
@floryst looks great, thanks!!

We are now done with alpha's, the new beta naming scheme sounds good to me.

I minor request inline.

thewtex (approved):
Thank you, @floryst !!

PR #574, opened 2022-05-31
ENH: Update http to https links

tbirdso:
Automated reformatting with https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/ApplyScriptToRemotes.sh

### DIFF
--- a/CMake/ExternalData.cmake
+++ b/CMake/ExternalData.cmake
@@ -28,7 +28,7 @@ For example:
  include(ExternalData)
  set(ExternalData_URL_TEMPLATES "file:///local/%(algo)/%(hash)"
                                 "file:////host/share/%(algo)/%(hash)"
-                                "http://data.org/%(algo)/%(hash)")
+                                "https://data.org/%(algo)/%(hash)")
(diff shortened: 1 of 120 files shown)


floryst:
LGTM.

Verifying https support for the rewritten links would be ideal, but realistically everyone should be on https. The one link I did test 404-ed (a bibliography link), but that's because of the page and not the https link.

floryst (approved):

tbirdso:
Thanks @floryst . I agree that ideally we would test each link individually, but as I am applying these updates to all ITK remote modules I don't have the manpower to go through each one. I'm relying primarily on CI to make sure nothing breaks in each build, if you or others find broken links in documentation it would be good to correct those in a subsequent fixup.

floryst:
Sure thing; I would much rather you not validate the links anyhow. :) I did a sanity check to ensure that code-related URLs are okay. The majority are non-code rewrites; those links are more tolerable to breakage.

PaulHax:
LGTM!

dzenanz (approved):

PR #584, opened 2022-06-21
ENH: return sorted file list when reading DICOM series

jadh4v:
The functions `readImageDICOMFileSeries` and `readImageDICOMArrayBufferSeries`
have an additional side effect that it sorts / orders the files
(particularly DICOM files) based on a pre-defined logic in `gdcm`.
However, this information is not related to the caller in any form.

Add `Array<string>` as the ordered list of filenames as additional output.

### DIFF
--- a/src/io/internal/pipelines/image/ReadDICOM/ReadImageDICOMFileSeries.cxx
+++ b/src/io/internal/pipelines/image/ReadDICOM/ReadImageDICOMFileSeries.cxx
@@ -30,6 +30,7 @@
 
 #include "itkPipeline.h"
 #include "itkOutputImage.h"
+#include "itkOutputTextStream.h"
 
 class CustomSerieHelper: public gdcm::SerieHelper
 {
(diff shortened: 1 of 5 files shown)


jadh4v:
@thewtex @floryst @sankhesh 
I am not fully certain of the API changes are a good move. We could also return this info in any other way like using Image object's metaData property. Unfortunately, the Instance UIDs for each file/slice are not available in the reader's pipeline unless we explicitly read the DICOM Tags. That is why I stuck with the filenames as the returned output. And the filenames don't seem suitable as metadata of the `Image`.

thewtex:
Hi @jadh4v thanks for working on this!

Why not add these as as `metadata` entries on the `Image`? Perhaps add both a `Filenames` and `InstanceUIDs` entry.

jadh4v:
I have added the returned ordered list of files as metadata to the output Image object rather than returning it as additional output for `readImageDICOMArrayBufferSeries`.

jadh4v:
@floryst @sankhesh please review

src/io/internal/pipelines/image/ReadDICOM/ReadImageDICOMFileSeries.cxx
```diff
@@ -257,11 +261,21 @@ int runPipeline(itk::wasm::Pipeline & pipeline, std::vector<std::string> & input
...
     reader->SetFileNames(inputFileNames);
+  }
+
+  // copy sorted filenames as additional output
+  if(!outputFilenamesOption->empty())
+  {
+    auto finalFileList = reader->GetFileNames();
+    for (auto f = finalFileList.begin(); f != finalFileList.end(); ++f)
+    {
+      outputFilenames.Get() << *f << ":";
```
floryst, line 276: Non-NTFS filenames can have a colon. Let's use a forward slash instead, since the majority of filesystems do not allow `/` in the filename.

jadh4v: Changed to use null char '\0' in place of ':'.

PR #588, opened 2022-07-01
feat(ReadImageDICOM): Return sorted file list when reading DICOM series

jadh4v:
This is to redo PR https://github.com/InsightSoftwareConsortium/itk-wasm/pull/584 with proper commit message format.

The functions readImageDICOMFileSeries and readImageDICOMArrayBufferSeries
have an additional side effect that it sorts / orders the files
(particularly DICOM files) based on a pre-defined logic in gdcm.
However, this information is not related to the caller in any form.

Add Array<string> as the ordered list of filenames as additional output.

### DIFF
--- a/src/core/Image.ts
+++ b/src/core/Image.ts
@@ -13,7 +13,7 @@ class Image {
 
   size: number[]
 
-  metadata: Record<string, string | number | number[] | number[][]>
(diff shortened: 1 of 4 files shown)


jadh4v:
cc: @floryst

floryst:
LGTM

PR #606, opened 2022-07-25
chore(commitlint): allow upper-case subject

floryst:
This allows for a subject line to start with an upper case.


thewtex:
@floryst thanks!

PR #611, opened 2022-08-15
fix(worker): Ensure worker promises are reused

cjh1:
When running task in a worker in "parallel" the access to the worker must be performed through the same worker promise otherwise requests and responses can get interleaved. This situation can occur if for example `Promise.all(...)` is used to run the tasks in the worker. To prevent this issue we attach the work promise to the worker, so it can be used again if the worker is reused.

### DIFF
--- a/src/core/internal/createWebWorkerPromise.ts
+++ b/src/core/internal/createWebWorkerPromise.ts
@@ -8,11 +8,24 @@ interface createWebWorkerPromiseResult {
   worker: Worker
 }
 
+interface itkWorker extends Worker {
+  workerPromise?: typeof WebworkerPromise
+}
+
 // Internal function to create a web worker promise
 async function createWebWorkerPromise (existingWorker: Worker | null): Promise<createWebWorkerPromiseResult> {
+  let workerPromise: typeof WebworkerPromise
   if (existingWorker != null) {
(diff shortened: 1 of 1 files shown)


src/core/internal/createWebWorkerPromise.ts
```diff
@@ -8,11 +8,26 @@ interface createWebWorkerPromiseResult {
   worker: Worker
```
thewtex, line 8: This should be `itkWorker` instead of `Worker`

thewtex (changes requested):
Thanks, @cjh1 !

Overall, lgtm.

A change requested in line.

src/core/internal/createWebWorkerPromise.ts
cjh1: We want to return the `Worker` interface. EWe only need it in the `itkWorker` form to do the assignment of the worker promise. The fact the `Worker` is augmented with the worker promise should be transparent to the user.

cjh1: I will fix up the lint issue

thewtex (approved):

cjh1:
Thanks @thewtex, do you think we could get a release with this change in soon?

PR #614, opened 2022-08-22
feat: expose version

FezVrasta:
The idea is to be able to do `import { version } from 'itk-wasm'` as it can be useful to know the version to use it in the build process to version the assets and invalidate caches.

### DIFF
--- a/src/itkConfig.ts
+++ b/src/itkConfig.ts
@@ -1,4 +1,4 @@
-import { version } from '.'
+import { version } from './index.js'
 
 const itkConfig = {
   webWorkersUrl: undefined,
--- a/src/itkConfig.ts
+++ b/src/itkConfig.ts
@@ -1,4 +1,4 @@
-const version = '1.0.0-b.18'
+import { version } from '.'
 
 const itkConfig = {
   webWorkersUrl: undefined,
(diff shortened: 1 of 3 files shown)


thewtex:
@FezVrasta thanks!

@floryst could you please take a look to make sure that the version is in sync with the generated semantic release?

CI also caught a syntax issue.

floryst:
Hard-coded versions get rewritten via [update-versions.cjs](https://github.com/InsightSoftwareConsortium/itk-wasm/blob/master/update-versions.cjs#L24), so that will also need to be updated to point to `core/index.ts`.

FezVrasta:
it should be okay now

src/itkConfig.ts
```diff
@@ -1,4 +1,4 @@
-const version = '1.0.0-b.18'
+import { version } from '.'
```
floryst, line 1: I think this should be `./index.js`. The JavaScript/TypeScript build is failing on this line.

floryst:
Once you rebase and squash the latest commit, we should be good to go!

floryst:
Forgot to mention, the `fix import` commit should either be squashed or conform to the semantic release angular conventions. Otherwise the PR checks won't pass.

thewtex (approved):
Awesome!!

PR #629, opened 2022-09-01
feat(structured-report-to-text): Initial pipeline addition

thewtex:


thewtex:
Merging so we can proceed to packaging

PR #633, opened 2022-09-06
feat(itk-wasm-cli): Update default Docker image for Pipeline updates

thewtex:


thewtex:
@sankhesh this will bring in #632 into the default Docker image

PR #635, opened 2022-09-07
ci(repository-url): Remove git+ protocol prefix

thewtex:
GitHub primarily supports the https protocol.


thewtex:
@floryst semantic-release is failing to authenticate in CI. This change is needed, but I am not sure it will fix the issue.

PR #646, opened 2022-09-14
refactor(itk-wasm-cli): Rename bindings subcommand to bindgen

thewtex:
More consistency with WASM community tooling.


thewtex:
@sankhesh @jadh4v name change heads up

PR #647, opened 2022-09-14
feat(itk-dicom): Support loading multi-component dicom data with gdcm

sankhesh:


sankhesh:
@thewtex Could you please review?

thewtex (changes requested):
@sankhesh thanks!

Since RGB pixels are usually only supported with `unsigned char`, let's only use `RGBAPixel` and `RGBPixel` for `unsigned char`. For the others with `case 4` and `case 3`, let's use `itk::Vector` as with `case 2`.

sankhesh:
Thanks @thewtex. I completely overlooked that `RGBPixels` typedefs to unsigned char only. While I fix it, is there a better way to do this without all the code duplication?

thewtex (approved):
:clap:

PR #649, opened 2022-09-23
Feature: add structured-report-to-html operation

jadh4v:
Add capability to read and render a dicom file structured report into an html document using implementation from DCMTK's `dsr2html` command line app.

### DIFF
--- a/src/core/internal/camelCase.ts
+++ b/src/core/internal/camelCase.ts
@@ -1,7 +1,17 @@
 function camelCase (kebobCase: string): string {
-  return kebobCase.replace(/-([a-z])/g, (kk) => {
-    return kk[1].toUpperCase()
-  })
+  // make any alphabets that follows '-' an uppercase character, and remove the corresponding hyphen
+  let cameledParam = kebobCase.replace(/-([a-z])/g, (kk) => {
+    return kk[1].toUpperCase();
+  });
+
+  // remove all non-alphanumeric characters
+  const outParam = cameledParam.replace(/([^0-9a-z])/ig, '')
+
+  // check if resulting string is empty
+  if(outParam === '') {
+    console.error(`Resulting string is empty.`)
+  }
+  return outParam
 }
 
 export default camelCase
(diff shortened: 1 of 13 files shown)


src/io/internal/pipelines/dicom/structured-report-to-html.cxx
```diff
@@ -0,0 +1,510 @@
+/*
+ *
+ *  Copyright (C) 2000-2016, OFFIS e.V.
```
jadh4v, line 3: @thewtex Do we want to add our own license header above or below this header? I assume the licenses be compatible?

thewtex: Yes, let's add it above

src/itk-wasm-cli.js
```diff
@@ -208,11 +208,21 @@ function run(wasmBinary, options) {
   }
 }
 
-
-function camelCase(kebobCase) {
-  return kebobCase.replace(/-([a-z])/g, (kk) => {
+function camelCase(param) {
+  // make any alphabets that follows '-' an uppercase character, and remove the corresponding hyphen
```
thewtex, line 212: please also update src/core/internal/camelCase.ts (we cannot import it here).

thewtex (changes requested):
@jadh4v excellent!

A few inline comments.

src/core/internal/camelCase.ts
```diff
@@ -1,7 +1,17 @@
...
+  let cameledParam = kebobCase.replace(/-([a-z])/g, (kk) => {
+    return kk[1].toUpperCase();
+  });
+
+  // remove '-' and '.' from the string
+  const outParam = cameledParam.replace(/-|\./g, '')
+
+  // check if resulting parameter name is alpha-numeric and starts with an alphabet
+  if(!outParam.match(/^([a-z])+([0-9a-z]+)$/i)) {
+    console.error(`Resulting parameter name is not alpha-numeric: ${outParam}`)
```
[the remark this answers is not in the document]
floryst: Instead of logging whether the parameter name is not valid, how about just stripping all non-valid chars to make outParam valid? I'd imagine most uses of this function will only have alphanumeric chars involved.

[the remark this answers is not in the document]
jadh4v: Ok, makes sense. I have changed the code to just replace all non-alphanumeric characters. I added to check for resulting empty strings.

thewtex (approved):
✨

PR #652, opened 2022-09-27
Feature: read dicom encapsulated pdf

jadh4v:
Add support for reading **Encapsulated PDF** DICOM SOP class (1.2.840.10008.5.1.4.1.1.104.1)

Output is return as a binary stream of the read pdf file.


thewtex (approved):
:100:

PR #658, opened 2022-09-30
feat(key-object-selection): support for rendering key object selection structured report

jadh4v:
Key object selection (KOS) structured report (SR) is similar to any other type of structured reports like radiation dose or comprehensive SR, except that a KOS has a list of referenced composite objects (images, waveforms, etc). These references are presented as a list of hyperlinks to the DICOM database files.

DCMTK renders these hyperlinks using a fixed literal ` http://localhost/dicom.cgi ` which is hard coded through a:
```c++
#define HTML_HYPERLINK_PREFIX_FOR_CGI
```
We replace this with a passed string argument to allow flexible hyperlinks based on the requirements of the calling application.

### DIFF
--- a/dist/dicom/src/structuredReportToHtml.ts
+++ b/dist/dicom/src/structuredReportToHtml.ts
@@ -1,6 +1,7 @@
 import {
   TextStream,
   InterfaceTypes,
+  PipelineInput,
   runPipeline
 } from 'itk-wasm'
 
@@ -23,7 +24,7 @@ async function structuredReportToHtml(
   const desiredOutputs = [
     { type: InterfaceTypes.TextStream },
   ]
-  const inputs = [
+  const inputs: [ PipelineInput ] = [
     { type: InterfaceTypes.BinaryFile, data: { data: dicomFile, path: "file0" }  },
   ]
 
@@ -80,13 +81,16 @@ async function structuredReportToHtml(
     args.push('--charset-require')
   }
   if (options.charsetAssume) {
-    args.push('--charset-assume', '1')
+    args.push('--charset-assume', options.charsetAssume.toString())
   }
   if (options.charsetCheckAll) {
     args.push('--charset-check-all')
   }
   if (options.convertToUtf8) {
     args.push('--convert-to-utf8')
+  }
+  if (options.urlPrefix) {
+    args.push('--url-prefix', options.urlPrefix.toString())
   }
   if (options.html32) {
     args.push('--html-3.2')
@@ -101,10 +105,14 @@ async function structuredReportToHtml(
     args.push('--add-document-type')
   }
   if (options.cssReference) {
-    args.push('--css-reference', '2')
+    const inputCountString = inputs.length.toString()
+    inputs.push({ type: InterfaceTypes.TextStream, data: { data: options.cssReference } })
+    args.push('--css-reference', inputCountString)
   }
   if (options.cssFile) {
-    args.push('--css-file', '3')
+    const inputFile = 'file' + inputs.length.toString()
+    inputs.push({ type: InterfaceTypes.TextFile, data: { data: options.cssFile, path: inputFile } })
+    args.push('--css-file', inputFile)
   }
   if (options.expandInline) {
     args.push('--expand-inline')
--- a/src/io/internal/pipelines/dicom/structured-report-to-html.cxx
+++ b/src/io/internal/pipelines/dicom/structured-report-to-html.cxx
@@ -243,7 +243,7 @@ int main(int argc, char * argv[])
   bool charsetRequire{false};
   pipeline.add_flag("--charset-require", charsetRequire,    "require declaration of ext. charset (default)");
   std::string charsetAssumeValue;
-  pipeline.add_option("--charset-assume", charsetAssumeValue, "[c]harset: string, assume charset c if no extended charset declared")->type_name("INPUT_TEXT_STREAM");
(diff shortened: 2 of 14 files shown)


src/io/internal/pipelines/dicom/structured-report-to-html.cxx
```diff
@@ -169,8 +170,20 @@ static OFCondition renderFile(STD_NAMESPACE ostream &out,
...
-            } else {
+                {
+                    if (urlPrefixForReferencedObjects.empty())
+                    {
+                        // this allows to set the default argument value: HTML_HYPERLINK_PREFIX_FOR_CGI
+                        result = dsrdoc->renderHTML(out, renderFlags, cssName);
+                    }
+                    else
+                    {
+                        result = dsrdoc->renderHTML(out, renderFlags, cssName, urlPrefixForReferencedObjects);
```
thewtex, line 181: Great

src/io/internal/pipelines/dicom/structured-report-to-html.cxx
```diff
@@ -260,10 +276,10 @@ int main(int argc, char * argv[])
 
   //   SubGroup("cascading style sheet (CSS), not with HTML 3.2:")
   std::string cssReferenceValue;
```
thewtex, line 278: While `--url-prefix`, `--charset-assume` would most easily be passed as small direct strings, `--css-reference` and `--css-file`, these will be larger and would be more easily passed on the command line as text files -> `itk::wasm::InputTextStream`.

src/itk-wasm-cli.js
```diff
@@ -240,6 +240,8 @@ const interfaceJsonTypeToTypeScriptType = new Map([
   ['INPUT_POLYDATA', 'PolyData'],
   ['OUTPUT_POLYDATA', 'PolyData'],
   ['BOOL', 'boolean'],
+  ['TEXT', 'string'],
+  ['INT', 'number'],
```
thewtex, line 244: Thanks

thewtex (changes requested):

src/io/internal/pipelines/dicom/structured-report-to-html.cxx
jadh4v: I have updated the code to handle `--css-reference` and `--css-file` as InputTextStream.

dist/dicom/src/structuredReportToHtml.ts
```diff
@@ -101,10 +105,12 @@ async function structuredReportToHtml(
     args.push('--add-document-type')
   }
   if (options.cssReference) {
-    args.push('--css-reference', '2')
+    args.push('--css-reference', '1')
+    inputs.push({ type: InterfaceTypes.TextStream, data: { data: options.cssReference } })
   }
   if (options.cssFile) {
-    args.push('--css-file', '3')
+    args.push('--css-file', '2')
```
thewtex, line 112: If `cssReference` was not passed, this would not be `2` -> let's keep track of the number of PipelineInput's that are used, increment the integer, then convert to string.

thewtex (changes requested):

dist/dicom/src/structuredReportToHtml.ts
jadh4v: Since we are using `inputs.push` for every such parameter, how about using `inputs.length` itself as the counter?

thewtex: Yes, good idea :zap: :bulb:

dist/dicom/dist/demo/app.js
```diff
@@ -1,4 +1,4 @@
-import { structuredReportToText } from '../itk-dicom.js'
+import { structuredReportToHtml } from '../itk-dicom.js'
```
thewtex, line 1: I can work on adding support for both in a followup.

For now this should either not be changed, or the test configuration also needs to updated:

https://github.com/InsightSoftwareConsortium/itk-wasm/blob/3efaf52f2fd2a42d0175e0510116d9025156940b/dist/dicom/cypress/e2e/structured_report_to_text.cy.ts

dist/dicom/src/structuredReportToHtml.ts
```diff
@@ -101,10 +105,14 @@ async function structuredReportToHtml(
...
   }
   if (options.cssReference) {
-    args.push('--css-reference', '2')
+    const inputCountString = ('TextStream').includes('File') ? 'file' + inputs.length.toString() : inputs.length.toString()
+    args.push('--css-reference', inputCountString)
+    inputs.push({ type: InterfaceTypes.TextStream, data: { data: options.cssReference, path: ('TextStream').includes('File') ? inputCountString : '' } })
   }
   if (options.cssFile) {
-    args.push('--css-file', '3')
+    const inputCountString = ('TextFile').includes('File') ? 'file' + inputs.length.toString() : inputs.length.toString()
```
thewtex, line 113: ` ('TextStream').includes('File')` and `('TextFile').includes('File')` are "static" -- could they be evaluated in `bindgen` execution so they do not need to be output and evaluated here?

thewtex (changes requested):
Nice! A few follow-ups inline.

dist/dicom/dist/demo/app.js
jadh4v: Ok, I can leave it for you to update later then. I will remove the corresponding commit.

thewtex (approved):
:sparkler: :100:

PR #659, opened 2022-10-05
feat(WASI): Support itk_wasm_delayed_start

thewtex:
Build as a Reactor but add a default export, defined by the empty string, that behaves like _start did before for command line execution via wasmtime.

  https://docs.rs/wasmtime/0.17.0/src/wasmtime/linker.rs.html#685


thewtex:
FF issues unrelated per #662

PR #660, opened 2022-10-06
ci(update-versions): Run during verifyReleaseCmd

thewtex:
According to:

  https://github.com/semantic-release/semantic-release/blob/master/docs/extending/plugins-list.md

The npm tarball is generated by the @semantic-release/npm plugin during the `prepare` step. So, bump the update version script execution to the `verifyReleaseCmd`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,7 +178,7 @@
       [
         "@semantic-release/exec",
         {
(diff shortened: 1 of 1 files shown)


sankhesh (approved):

thewtex:
FireFox failures unrelated re #661

PR #668, opened 2022-10-11
fix: Run update-versions during prepublishOnly npm step

thewtex:
Get the version from package.json, which should be set by semantic-release.

### DIFF
--- a/src/itkConfigDevelopment.ts
+++ b/src/itkConfigDevelopment.ts
@@ -1,4 +1,4 @@
-const version = '1.0.0-b.18'
+const version = '0.0.0-semantically-released'
 
 const itkConfig = {
   webWorkersUrl: undefined,
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "itk-wasm",
-  "version": "1.0.0-b.18",
+  "version": "0.0.0-semantically-released",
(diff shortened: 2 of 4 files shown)


src/itkConfigDevelopment.ts
```diff
@@ -1,4 +1,4 @@
-const version = '1.0.0-b.18'
+const version = '1.0.0-b.38'
```
FezVrasta, line 1: Would it make sense to insert a placeholder here instead? And have the build script find/replace it?

```suggestion
const version = '<%PKG_VERSION%>'
```

floryst: I prefer this approach, since I don't think there's any bit of internal code that actually cares about the format of the `version` variable. Either a regular string.replace call works fine, or the use of lodash templates.

thewtex: Yes, agreed.

floryst:
@thewtex Were you able to verify this will work with a dry run?

thewtex:
@floryst Yes, locally with `npm publish --dry-run`. I am not sure how to test with semantic-release.

floryst:
I would presume that semantic-release will have written out the correct version by the prepublish step.

floryst (approved):

Issue #671, opened 2022-10-14
Allow to specify scalars data type?


FezVrasta:
As part of some performance optimizations work we are doing, we noticed that a huge bottleneck is located at the vtk.js `updateArrayDataType` helper (of Rendering/OpenGL/Texture).

The issue is that `image.getPointData().getScalars()` can return any kind of typed array, while the webgl logic requires a Float32Array. Any time vtk needs to render something, it needs to convert the scalars to the expected format, and this produces many garbage collection events and general slowness due to the size of the scalars data.

If itk-wasm provided an option to pre-convert the scalars in the Float32Array type, we could avoid all the above.

Do you think this is something that you could add to this lib?

PR #678, opened 2022-10-27
feat(castImage): Function to cast to pixelType, componentType

thewtex:


thewtex:
Towards #671

cypress/e2e/core/castImage.cy.ts
```diff
@@ -0,0 +1,148 @@
...
+
+      const baseline = inputImage
+      baseline.imageType.componentType = itk.IntTypes.UInt64
+      baseline.data = new BigUint64Array(baseline.data.length)
+      baseline.data.fill(7n)
+
+      compareImageToBaseline(itk, outputImage, baseline)
+    })
+  })
+})
```
jadh4v, line 148: Need a newline at eof?

src/core/castImage.ts
```diff
@@ -0,0 +1,158 @@
...
+ * @param {Image} image - The input image
+ * @param {CastImageOptions} options - specify the componentType and/or pixelType of the output
+ */
+function castImage (inputImage: Image, options: CastImageOptions): Image {
+  const outputImageType = { ...inputImage.imageType }
+
+  if (typeof options.pixelType !== 'undefined') {
+    outputImageType.pixelType = options.pixelType
+    if (options.pixelType === PixelTypes.Scalar && outputImageType.components !== 1) {
+      throw new Error('Cannot cast multi-component image to a scalar image')
```
jadh4v, line 19: Would it also make sense to check data range and warn about data loss due to narrowing conversions?

src/core/castImage.ts
```diff
@@ -0,0 +1,158 @@
...
+    outputImageType.componentType = options.componentType
+  }
+
+  const outputImage = new Image(outputImageType)
+
+  outputImage.name = inputImage.name
+  outputImage.origin = Array.from(inputImage.origin)
+  outputImage.spacing = Array.from(inputImage.spacing)
+  outputImage.direction = inputImage.direction.slice()
+  outputImage.size = Array.from(inputImage.size)
```
jadh4v, line 32: Could be better to do a deep copy of `inputImage` and then replace `outputImage.componentType` and `outputImage.data` but perhaps typescript wouldn't allow it. It feels like using a copy constructor would be better here as adding more properties to the Image type in the future will cause problems.
Alternatively, at least also copy `Image.metadata` which is being used for storing filenames of the dicom series, etc.

src/core/castImage.ts
```diff
@@ -0,0 +1,158 @@
...
+            case FloatTypes.Float32:
+              outputImage.data = new Float32Array(inputImage.data)
+              break
+            case FloatTypes.Float64:
+              outputImage.data = new Float64Array(inputImage.data)
+              break
+            case IntTypes.UInt64:
+              outputImage.data = new BigUint64Array(inputImage.data.length)
+              for (let idx = 0; idx < outputImage.data.length; idx++) {
+                outputImage.data[idx] = BigInt(inputImage.data[idx])
```
jadh4v, line 73: Should we use `BigInt.asUintN(64, num)` here?

src/core/castImage.ts
thewtex: Worth considering, but in the trade-off of developer safety nets and performance, we do not want to iterate through the data here to compute the range.

src/core/castImage.ts
thewtex: Yes, sharing code when possible makes sense. We have to skip a full copy constructor because we do not want an extra pixel array buffer allocation and copy. I've added a copy of the `metadata`.

src/core/castImage.ts
thewtex: Yes!

jadh4v (approved):
LGTM

floryst:
LGTM

PR #682, opened 2022-11-04
Image IO componentType, pixelType options

thewtex:
- `pixelType` and `componentType` options to image IO functions
- Same option object conventions as other pipeline functions. Existing options arguments still supported through backwards compatibility code.


thewtex:
Closes #671

src/io/readImageDICOMArrayBufferSeries.ts
```diff
@@ -78,11 +80,29 @@ const seriesBlockSize = 8
...
+  let fileNames: undefined | string[]
+  if (typeof options === 'boolean') {
+    // Backwards compatibility
+    singleSortedSeries = options
+  }
+  if (typeof fileNamesBackwardsCompatibility !== 'undefined') {
+    fileNames = fileNamesBackwardsCompatibility
+  }
+  if (typeof options === 'object') {
+    if (typeof options.singleSortedSeries !== 'undefined') {
```
sankhesh, line 96: Should `options.singleSortedSeries` type be enforced as boolean?

sankhesh (approved):
LGTM overall except that one question

src/io/readImageDICOMArrayBufferSeries.ts
thewtex: Yes, this is already enforced via Typescript at compile time. This helps remove overhead of a run-time check.

PR #691, opened 2022-11-14
Metadata support

thewtex:


thewtex:
Note a breaking change of `image.metadata` from an Object to a Map.

sankhesh (approved):
@thewtex  Being able to serialize the metadata would be really useful.

thewtex:
Note that this is going to produce a dictionary of DICOM tags for multi-frame or single frame DICOM and provide the NIFTI header information for NIFTI images.

PR #694, opened 2022-11-17
docs: Add BREAKING_CHANGES.md to track beta breaking changes

thewtex:
This manually maintained files help document breaking changes in the 1.0 beta pre-release phase.

Breaking changes also continue to be noted with `BREAKING CHANGE:` in commit messages according to the conventional-changelog angular specification.

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,0 +1,15 @@
+# itk-wasm breaking changes
+
+This is a log of itk-wasm breaking changes in the 1.0 beta pre-release. We work to
+limit any breaking changes to when they are needed, provide backwards
+compatible interfaces when possible, and document breaking changes. However,
+for long-term improvement of the package, breaking changes are occasionally
+needed. These changes are noted here.
+
+For a log and guide when transitioning from itk.js to itk-wasm,
+please see [the migration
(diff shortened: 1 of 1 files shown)


floryst (approved):
LGTM. @FezVrasta please add any breaking changes that you've noted. 

Separately, we should be more cognizant about marking breaking changes as such. Browsing the CHANGELOG.md, I don't see any recent breaking changes, which means we didn't record them.

thewtex:
> Separately, we should be more cognizant about marking breaking changes as such. Browsing the CHANGELOG.md, I don't see any recent breaking changes, which means we didn't record them.

Yes, we should be as cognizant as possible about marking breaking changes.

Note that in `beta` semantic-release does not point out BREAKING_CHANGES in the CHANGELOG.md file. Once we get to 1.0.0, that will also be improved.

floryst:
Oh huh, so it doesn't. There are breaking change commits that don't show up in the beta release notes. Didn't realize that extended to the release notes generation as well.

jadh4v:
Does `BREAKING CHANGE:` need to be in the commit title, or can be anywhere in the commit msg body?

jadh4v (approved):

floryst:
> Does `BREAKING CHANGE:` need to be in the commit title, or can be anywhere in the commit msg body?

It needs to be in the message body.

PR #733, opened 2023-01-25
docs(docsify): Add redirects for shared links

thewtex:
Workaround if we cannot use redirect rules re:
https://github.com/FleekHQ/fleek-docs/issues/154 .

### DIFF
--- a/docs/api/index.html
+++ b/docs/api/index.html
@@ -1,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="refresh" content="0; url='https://wasm.itk.org/'" />
+  </head>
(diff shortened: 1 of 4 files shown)


thewtex:
@finetjul this is not perfect, but it will help.

finetjul:
I did not see a difference, e.g. [https://wasm.itk.org/tutorial/hello_world?id=introduction](https://wasm.itk.org/tutorial/hello_world?id=introduction)

thewtex:
@finetjul yes, this only makes https://wasm.itk.org/tutorial/ redirect to the home page. Tracking issue here: https://github.com/InsightSoftwareConsortium/itk-wasm/issues/742

PR #735, opened 2023-01-25
build(semantic-release): Bump packages to latest versions

thewtex:
Also install @semantic-release/github, which is used but explicitly in the dependencies.

Attempt to avoid potential GitHub API over usage.

Suggested-by: Zach Mullen <zach.mullen@kitware.com>


floryst:
LGTM!

PR #738, opened 2023-01-27
feat(bindgen): Default to itk-wasm package asset configuration

thewtex:
At runtime, we use in order of preference:

1. Package asset configuration
2. itk-wasm asset configuration
3. jsDeliver

### DIFF
--- a/packages/compress-stringify/typescript/src/pipeline-worker-url.ts
+++ b/packages/compress-stringify/typescript/src/pipeline-worker-url.ts
@@ -1,10 +1,21 @@
+// @ts-ignore: TS2305: Module '"itk-wasm"' has no exported member 'getPipelineWorkerUrl'.
+import { getPipelineWorkerUrl as itkWasmGetPipelineWorkerUrl } from 'itk-wasm'
 import packageJson from '../package.json'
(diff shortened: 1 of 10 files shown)


packages/compress-stringify/typescript/src/pipeline-worker-url.ts
```diff
@@ -1,10 +1,21 @@
+// @ts-ignore: TS2305: Module '"itk-wasm"' has no exported member 'getPipelineWorkerUrl'.
+import { getPipelineWorkerUrl as itkWasmGetPipelineWorkerUrl } from 'itk-wasm'
 import packageJson from '../package.json'
-let pipelineWorkerUrl: string | URL | null = `https://cdn.jsdelivr.net/npm/itk-compress-stringify@${packageJson.version}/dist/web-workers/pipeline.worker.js`
+
+let pipelineWorkerUrl: string | URL | null | undefined
+let defaultPipelineWorkerUrl: string | URL | null = `https://cdn.jsdelivr.net/npm/itk-compress-stringify@${packageJson.version}/dist/web-workers/pipeline.worker.js`
```
floryst, line 6: `const` if not going to modify? Also, doesn't need to have the extra union types.

PR #740, opened 2023-01-27
refac(itk-dicom): refactor itk-dicom as package

jadh4v:


jadh4v:
@thewtex Should this line be removed to updated to the new path?
https://github.com/jadh4v/itk-wasm/blob/ENH-migrate-dicom-module-to-separate-package/src/CMakeLists.txt#L44

packages/dicom/typescript/index.html
```diff
@@ -77,7 +76,7 @@
       </section>
     </sp-theme>
 
-    <script type="module" src="./dist/itk-dicom.js"></script>
-    <script type="module" src="./dist/demo/app.js"></script>
+    <script type="module" src="./dist/bundles/itk-dicom.js"></script>
+    <script type="module" src="./test/browser/app.ts"></script>
```
jadh4v, line 80: We are changing `dist/demo/app.js` to `test/browser/app.ts` but this is a demo and not an automated browser test. Is this really correct?

thewtex: It is both a demo and the basis for tests with Cypress.

thewtex (approved):

PR #741, opened 2023-01-27
Migrate image io pkg

floryst:


thewtex:
rebasing on `main` and re-starting migration

thewtex:
Building on @floryst 's work here and adding @aylward 's suggestion, we support asking all ITK image-io's for `information-only`, calling the IO's `ReadImageInformation()`, to get the metadata separately.

This can be exposed in `itk-viewer/io` in an `ItkImageIoMultiscaleSpatialImage`, similar to a `DicomSeriesMultiscaleSpatialImage` suggested by @floryst and the `ZarrMultiscaleSpatialImage` implemented by @PaulHax .

PR #746, opened 2023-01-31
docs(Examples): Add different-input-types examples

thewtex:
Demonstrates how to create a pipeline that accepts two different input image types.


examples/different-input-types/different-input-types.cxx
```diff
@@ -0,0 +1,117 @@
...
+
+int main( int argc, char * argv[] )
+{
+  // Create the pipeline for parsing arguments. Provide a description.
+  itk::wasm::Pipeline pipeline("different-input-types", "An itk-wasm pipeline example that demonstrates accepting different input types", argc, argv);
+
+  return itk::wasm::SupportInputImageTypes<InputImagePipelineFunctor,
+   uint8_t,
+   uint16_t>
+  ::Dimensions<2U>("input-image", pipeline);
```
PaulHax, line 116: If I add a third Dimension
`::Dimensions<2U, 3U>("input-image", pipeline);`
I get compile errors.

How can I narrow down the TImage and TLabelImage dimension to just be matching?

PaulHax (approved):
Very helpful.

The `wasi-build` directory is not in `.gitignore`.

examples/different-input-types/different-input-types.cxx
thewtex: Good point! Let's fix the second image dimension to the first image dimension as a best practice. Also adding .gitignore.

PR #764, opened 2023-02-14
build(bindgen): Add missing build resources

thewtex:
In testing with ITKMeshToPolyData `@types/node` required for 'path' in the typescript build. `debug` for the vite dev build.


PaulHax (approved):
Builds good with ITKMeshToPolyData.  Is a Vite warning on `npm run start` for the optional dependency `supports-color`.  Toss that one in too?  I'm not sure how to disable the Vite warning.

PR #771, opened 2023-02-24
build(python): Do not depend on wasmer with emscripten

thewtex:


thewtex:
@bnmajor this will address the JupyterLite dependency -- pushed to PyPI.

bnmajor:
@thewtex Awesome, thanks!

PR #783, opened 2023-03-07
fix(itkWasmImageToImageFilter): use PixelType instead of IOPixelType

PaulHax:
For VariableLengthVector support


thewtex (approved):
:100: @PaulHax thanks!!

PR #798, opened 2023-03-21
fix(cli): Fix the cmakeargs for windows

### LINKED ISSUE #797: itk-wasm-cli doesn't work on windows with more than one cmake arg
Currently if on windows itk-wasm-cli incorrectly constructs a string for the cmake arguments passed to the build command after `"--"`. The cmakeargs variable is an Array of strings. It is added using the plus operator to another string, when adding an Array to a string the default behavior is to call Array.toString() which will join the strings with a comma. This results in cmake not recognizing the arguments correctly as it expects them to be delimited with a space. A potential fix is to call `cmakeArgs.join(' ')` instead.

alexDrinkwater:
fixes #797

### DIFF
--- a/src/itk-wasm-cli.js
+++ b/src/itk-wasm-cli.js
@@ -117,7 +117,7 @@ function build(options) {
   }
   if(process.platform === "win32"){
     var dockerBuild = spawnSync('"C:\\Program Files\\Git\\bin\\sh.exe"',
-      ["--login", "-i", "-c", `"${buildDir}/itk-wasm-build-env web-build ${buildDir} ` + cmakeArgs + '"'], {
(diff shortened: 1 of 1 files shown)


jspanchu:
FYI, @thewtex

thewtex (approved):
@alexDrinkwater @jspanchu thank you!!

alexDrinkwater:
@thewtex Not sure of how you manage your releases, but when is this expected to be merged?

thewtex:
@alexDrinkwater this took a little longer to be merged since CI needed to be approved for a first-time contributor on the repository.

Merging now, and this CD will deploy this to a new version after the CI builds complete.

PR #830, opened 2023-05-09
fix(package): add types entry to exports

PaulHax:
Allows Typescript project to use `nodenext` module resolution strategy.

```json
"compilerOptions": {
    "moduleResolution": "nodenext"
}
```
More info:
https://moonrepo.dev/docs/guides/javascript/typescript-project-refs#enabling-importsexports-resolution

and of course:
https://nodejs.org/api/packages.html#community-conditions-definitions

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


thewtex (approved):
:100:

PR #859, opened 2023-06-12
Apply presentation description null

thewtex:


packages/dicom/apply-presentation-state-to-image.cxx
```diff
@@ -513,13 +549,16 @@ int GenerateOutputImage(OutputImageType& outputImage, const unsigned long width,
   importFilter->SetImportPointer((PixelType*)pixelData, numberOfPixels, false);
   importFilter->Update();
 
+  typename OutputImageType::Pointer copy;
+  itk::ImageAlgorithm::Copy(importFilter->GetOutput(), copy.GetPointer(), region, region);
+
   // set as output image
-  outputImage.Set(importFilter->GetOutput());
+  outputImage = copy;
```
jadh4v, line 556: Do we want to use the call to `itk::ImageAlgorithm::Copy` in place of the previous `outputImage.Set()` call? The crash appears to be due to this change.

thewtex: @jadh4v yes, it should have been `.Set` here, but I was trying this copy because I was not sure if there was an issue with DCMTK's ownership / lifetime of the pixel buffer. But, tests are passing locally. Thanks for your help with this!

PR #861, opened 2023-06-14
fix(dicom-node): export location

jadh4v:
Update export location to match package installation.

Fix for error:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
'...node_modules\@itk-wasm\dicom\dist\bundles\dicom.node.js'
imported from ...
```


thewtex (approved):
Yes!

PR #865, opened 2023-06-21
Cxa thread atexit shim

thewtex:


src/cxaThreadAtExitShim.cxx
```diff
@@ -0,0 +1,38 @@
...
+
+  using Dtor = void(*)(void*);
+
+}
+
+extern "C" {
+
+int __cxa_thread_atexit(__cxxabiv1::Dtor dtor, void* obj, void* dso_symbol) throw()
+{
+  // Do nothing
```
N-Dekker, line 33: Should it not `return` anything?

thewtex: We could have it return an `int`.

N-Dekker: OK, but then it should have a `return` statement, right?

thewtex: Yes, it now returns -1, which seems to be what is used when the function does not run as expected.

N-Dekker: Thanks for addressing my comment 👍

PR #886, opened 2023-08-09
fix(readImageDicomFileSeries): await if file objects are passed

jadh4v:
An await is required if File objects are passed to `readImageDicomFileSeries()`. This is not caught in current cypress tests, since the test calls `readImageDicomFileSeries` with ArrayBuffer objects which doesn't require await since already in memory).

Error details:
```
CLI::ArgumentMismatch: --input-images: 1 required INPUT_BINARY_FILE:FILE missing
```

This fix is related to the following PR for VolView:
https://github.com/Kitware/VolView/pull/385
see: https://github.com/Kitware/VolView/pull/385#discussion_r1286418006

### DIFF
--- a/packages/dicom/typescript/src/read-image-dicom-file-series.ts
+++ b/packages/dicom/typescript/src/read-image-dicom-file-series.ts
@@ -34,14 +34,14 @@ async function readImageDicomFileSeries(
     throw new Error('"input-images" option must have a length > 1')
   }
 
-  options.inputImages.forEach(async (value) => {
+  await Promise.all(options.inputImages.map(async (value) => {
     let valueFile = value
     if (value instanceof File) {
       const valueBuffer = await value.arrayBuffer()
       valueFile = { path: value.name, data: new Uint8Array(valueBuffer) }
     }
     inputs.push(valueFile as BinaryFile)
-  })
+  }))
 
   if (options.singleSortedSeries) {
     const taskArgsArray = []
--- a/src/bindgen/typescript/function-module.js
+++ b/src/bindgen/typescript/function-module.js
@@ -279,7 +279,7 @@ function functionModule (srcOutputDir, forNode, interfaceJson, modulePascalCase,
       if (forNode) {
         functionContent += `    options.${camel}.forEach((value) => {\n`
       } else {
-        functionContent += `    options.${camel}.forEach(async (value) => {\n`
+        functionContent += `    await Promise.all(options.${camel}.map(async (value) => {\n`
       }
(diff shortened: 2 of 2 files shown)


packages/dicom/typescript/src/read-image-dicom-file-series.ts
```diff
@@ -34,14 +34,14 @@ async function readImageDicomFileSeries(
     throw new Error('"input-images" option must have a length > 1')
   }
 
-  options.inputImages.forEach(async (value) => {
+  await Promise.all(options.inputImages.map(async (value) => {
```
PaulHax, line 37: Cool!  

Should we just return the transformed image instead of pushing into array?

`const inputs = await Promise.all(`

thewtex: good idea, but `inputs` can also have content from other input's in the general case.

thewtex (changes requested):
@jadh4v thank you!!

Could the change also please be made in the generator:

https://insightsoftwareconsortium.github.io/GetYourBrainTogether/HCK02_2023_Allen_Institute_Hybrid/Tutorials/visualization/

packages/dicom/typescript/src/read-image-dicom-file-series.ts
jadh4v: ahh, I always make this mistake. For future, would it make sense to have some file naming convention to indicate generated files? like `*.gen.ts, *.gen.js` etc?

jadh4v:
> @jadh4v thank you!!
> 
> Could the change also please be made in the generator:
> 
> https://github.com/InsightSoftwareConsortium/itk-wasm/blob/c4f1b490019bdf3906d52f1edef3a7c4a7613ac4/src/bindgen/typescript/function-module.js#L282

ahh, I always make this mistake. For future, would it make sense to have some file naming convention to indicate generated files? like *.gen.ts, *.gen.js etc? Because this is going to be a repeated error on new developers side who don't realize it.

thewtex:
@jadh4v yes, good point. I just added a comment at the top of generated files accordingly:

https://github.com/InsightSoftwareConsortium/itk-wasm/blob/c4f1b490019bdf3906d52f1edef3a7c4a7613ac4/packages/dicom/typescript/src/apply-presentation-state-to-image.ts#L1

But `read-image-dicom-file-series.ts` is an exception -- it started from generated state, but the custom parallel performance behavior was added, so the comment was removed.

thewtex (approved):
:1st_place_medal:

thewtex:
Rebasing on `main` for CI

Issue #608, opened 2022-08-11
Some 3D DICOM files fail to load


timmevandermeer:
I use itk.wasm to load DICOM files into a browser application. This typically concerns 3D DICOM files of between 100 and 400 mb in size. For most files this works fine using `readImageBlob`. However, for some files the following error occurs:

<details>
<summary>Aborted() error in GDCMImageIOReadImage.umd.js:9</summary>

```js
Aborted()
Module.printErr | @ | GDCMImageIOReadImage.umd.js:9
-- | -- | --
  | abort | @ | GDCMImageIOReadImage.umd.js:9
  | _abort | @ | GDCMImageIOReadImage.umd.js:9
  | $func42 | @ | 00ad2fc6:0x2715
  | $func374 | @ | 00ad2fc6:0x22cc3
  | $func99 | @ | 00ad2fc6:0x6629
  | $func2036 | @ | 00ad2fc6:0x127c84
  | $func2042 | @ | 00ad2fc6:0x1280db
  | $func54 | @ | 00ad2fc6:0x4787
  | $func444 | @ | 00ad2fc6:0x35331
  | $func1265 | @ | 00ad2fc6:0xcf792
  | $func576 | @ | 00ad2fc6:0x469c1
  | $func260 | @ | 00ad2fc6:0x144b6
  | $func940 | @ | 00ad2fc6:0x8a904
  | $func1730 | @ | 00ad2fc6:0x113e7c
  | $func703 | @ | 00ad2fc6:0x605e0
  | $R | @ | 00ad2fc6:0x1651ad
  | Module._main | @ | GDCMImageIOReadImage.umd.js:9
  | callMain | @ | GDCMImageIOReadImage.umd.js:9
  | M_ | @ | 6012291f-afcb-4805-8260-6c8b6a589776:3
  | (anonymous) | @ | 6012291f-afcb-4805-8260-6c8b6a589776:3
  | p | @ | 6012291f-afcb-4805-8260-6c8b6a589776:3
  | (anonymous) | @ | 6012291f-afcb-4805-8260-6c8b6a589776:3
  | (anonymous) | @ | 6012291f-afcb-4805-8260-6c8b6a589776:3
  | e | @ | 6012291f-afcb-4805-8260-6c8b6a589776:1
  | u | @ | 6012291f-afcb-4805-8260-6c8b6a589776:1
  | (anonymous) | @ | 6012291f-afcb-4805-8260-6c8b6a589776:1
[...]

PR #895, opened 2023-08-15
feat: grow WASM memory to 4GB limit

PaulHax:
Rases WASM memory cap from 2GB to 4GB.  

Copes with this error when running pipelines.  
```
Cannot enlarge memory, asked to go up to 2248392704 bytes, but the limit is 2147483648 bytes!
```

Useful for trying to load large DICOM series with `readImageDicomFileSeries`  

Copes with #608


thewtex (approved):
Awesome!

thewtex:
Pushed to DockerHub with tag `20230815-369e670d`

PR #923, opened 2023-08-31
fix(dicom): typo in node export

floryst:
The output file is `dicom-node.js` instead of `dicom.node.js`.


thewtex (approved):
:+1: thanks!

PR #907, opened 2023-08-22
ci(actions): Fix GitHub actions workflow warnings linked to `Node.js`

jhlegarreta:
Fix GitHub actions workflow warnings linked to `Node.js`: bump `actions/checkout` and `actions/setup-node` to `v3`.

Fixes:
```
The following actions uses node12 which is deprecated and will be forced to run on node16:
actions/checkout@v2, actions/setup-node@v2.
For more info:
https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```

raised for example in:
https://github.com/InsightSoftwareConsortium/itk-wasm/actions/runs/5943988559

PR #934, opened 2023-09-16
ci: make ubuntu versions consistent across workflow file jobs

jhlegarreta:
- ci: make ubuntu versions consistent across workflow file jobs
- ci: use GitHub Action to free up disk space

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -5,7 +5,7 @@ on: [push,pull_request]
 jobs:
   test-node-example:
     name: node-js
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     defaults:
       run:
         working-directory: ./examples/node-js
@@ -116,7 +116,7 @@ jobs:
 
   build-test-hello-pipeline-example:
     name: hello-pipeline build test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     defaults:
       run:
@@ -160,7 +160,7 @@ jobs:
 
   build-test-inputs-outputs-example:
     name: inputs-outputs
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     defaults:
       run:
@@ -201,7 +201,7 @@ jobs:
 
   test-umd-example:
     name: umd
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v3
@@ -230,7 +230,7 @@ jobs:
 
   test-webpack-example:
     name: webpack
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v3
@@ -266,7 +266,7 @@ jobs:
 
   test-vite-example:
     name: vite
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


jhlegarreta:
Cross-referencing https://github.com/InsightSoftwareConsortium/itk-wasm/pull/907#issuecomment-1689780836.

Notes about the commit message:
- Used [commitizen](https://github.com/commitizen-tools/commitizen) to create the commit using `cz commit`. Hope this is the correct way.
- `cz commit` recommended to keep the subject to all lowercases. Let me know if we follow that here.
- `cz commit` does not seem to enforce a new line at e.g. 72 characters. Let me know if a line length limit is used here, and if so, how to configure that.

Thanks.

thewtex (approved):

thewtex (changes requested):
CI catches updates needed in the "Free up disk space" sections of the CI configurations

thewtex:
Thank you Jon Haitz @jhlegarreta !

It looks like the clean-up disk space steps need to be updated for Ubuntu 22.04.


> Used commitizen to create the commit using cz commit. Hope this is the correct way.

That works.

> cz commit recommended to keep the subject to all lowercases. Let me know if we follow that here.

Currently we support all lowercase but also starting the subject with a capital, e.g. `ci: make ubuntu versions consistent` or `ci: Make ubuntu versions consistent`.

> cz commit does not seem to enforce a new line at e.g. 72 characters. Let me know if a line length limit is used here, and if so, how to configure that.

There is a line length of 300, configured here: https://github.com/InsightSoftwareConsortium/itk-wasm/blob/b1289e36d0b2483c8687d9325fb16ed604be4b15/package.json#L242-L245

for pasting long error messages. CI will fail if that is exceeded.

jhlegarreta:
> It looks like the clean-up disk space steps need to be updated for Ubuntu 22.04.

See if 0558703 makes the trick.

> Currently we support all lowercase but also starting the subject with a capital, e.g. ci: make ubuntu versions consistent or ci: Make ubuntu versions consistent.

OK.

> There is a line length of 300, configured here:

OK.

jhlegarreta:
> See if 0558703 makes the trick.

The disk space issue is still not solved:
https://github.com/InsightSoftwareConsortium/itk-wasm/actions/runs/6229582807/job/16908321942?pr=934

thewtex:
Perhaps an `apt-get update` is needed to update mirror contents?

thewtex:
Seems to working in the others, but the `Javascript, Typescript` looks to need updates.

jhlegarreta:
Reworded my commit message to honor its changes, as freeing disk space is now addressed in a separate commit.

.github/workflows/examples.yml
```diff
@@ -206,20 +159,14 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: 'Free up disk space'
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@main
+
+    - name: Install cypress deps
```
jhlegarreta, line 165: Should this be put into a separate commit/PR ? I'd say that some tests were failing after this addition.

thewtex: Yes, I added another commit to try to get this green for integration.

thewtex (approved):
Thanks again @jhlegarreta for your awesome work on this!

The remaining failing CI check was addressed in `main` by https://github.com/InsightSoftwareConsortium/itk-wasm/commit/8a9925b3647993e0074d486877bd02492c4f5c93

PR #936, opened 2023-09-16
build(cmake): set `download_extract_timestamp` to `true`

jhlegarreta:
Set `DOWNLOAD_EXTRACT_TIMESTAMP` to `TRUE`: the timestamps of the extracted files will reflect the timestamps in the archive.

Fixes:
```
CMake Warning (dev) at D:/a/_temp/-2103337693/cmake-3.24.2-windows-x86_64/share/cmake-3.24/Modules/ExternalProject.cmake:3074 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
```

raised for exmaple in:
https://open.cdash.org/build/8983547/configure

Related documentation:
https://cmake.org/cmake/help/latest/module/ExternalProject.html?highlight=download_extract_timestamp https://cmake.org/cmake/help/latest/policy/CMP0135.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,12 @@ project(WebAssemblyInterface)
 
 set(CMAKE_CXX_STANDARD 17)
 
+if(${CMAKE_VERSION} VERSION_LESS 3.24)
+  set(download_extract_timestamp_flag)
+else()
+  set(download_extract_timestamp_flag DOWNLOAD_EXTRACT_TIMESTAMP TRUE)
+endif()
+
 set(WebAssemblyInterface_LIBRARIES WebAssemblyInterface)
 
 option(BUILD_ITK_WASM_IO_MODULES "Build the itk-wasm ImageIO's and MeshIO's" OFF)


jhlegarreta:
A number of other CMake warnings are present in the GHA workflow log:
https://github.com/InsightSoftwareConsortium/itk-wasm/actions/runs/5921543455/job/16054231396#step:7:132

But are not reported to CDash. Not sure why.

They may well deserve some attention.

src/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+if(${CMAKE_VERSION} VERSION_LESS 3.24)
+  set(download_extract_timestamp_flag)
+else()
+  set(download_extract_timestamp_flag DOWNLOAD_EXTRACT_TIMESTAMP TRUE)
+endif()
```
jhlegarreta, line 5: Not sure if this should go here.

CMakeLists.txt
```diff
@@ -3,6 +3,12 @@ project(WebAssemblyInterface)
 
 set(CMAKE_CXX_STANDARD 17)
 
+if(${CMAKE_VERSION} VERSION_LESS 3.24)
+  set(download_extract_timestamp_flag)
+else()
+  set(download_extract_timestamp_flag DOWNLOAD_EXTRACT_TIMESTAMP TRUE)
+endif()
```
jhlegarreta, line 10: Not sure if this is the appropriate place within the file.

jhlegarreta: Resolving then.

thewtex:
Failing CI check addressed in https://github.com/InsightSoftwareConsortium/itk-wasm/pull/940.

CMakeLists.txt
jhlegarreta: Looks like it did not work:
https://github.com/InsightSoftwareConsortium/ITKElastix/pull/248#discussion_r1329355073

PR #937, opened 2023-09-16
docs: add contributing documentation file

jhlegarreta:
Add contributing documentation file: cross-reference the contributing
section of the `hacking_itk_wasm.md` file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+Please see [our contributing documentation](docs/development/hacking_itk_wasm.md#contributing) for more information.


CONTRIBUTING.md
```diff
@@ -0,0 +1 @@
+{{docs/development/hacking_itk_wasm.md#contributing}}
```
jhlegarreta, line 1: Not sure at all that this will work.

thewtex: Since GitHub does not render it, what you think about something like,

```
Please see [our contributing documentation](https://wasm.itk.org/en/latest/development/hacking_itk_wasm.html) for more information.
```

?

jhlegarreta: I made a few attempts in a test repository to see if the transclusion worked but did not succeed. I do not like the cross-reference solution very much, but I adopted it: instead of using the HTML link, I used a relative link directly to the section of interest in the Markdown file at issue.

thewtex (approved):
@jhlegarreta thank you!!

PR #964, opened 2023-10-11
Stack save restore

thewtex:


floryst (approved):
LGTM! Thanks!

thewtex:
I think the test failures here are related to `latest` images not being those expected -- the tests pass locally.

Issue #476, opened 2022-02-07
Provide support for Webpack 5 workers


FezVrasta:
Webpack 5 supports Web Woekers natively:

https://webpack.js.org/guides/web-workers/

This library could leverage this to provide a more ergonomic system to consume the library that doesn't require to manually copy the webworker files to the public folder.

FezVrasta:
Thanks, does it work for image-io too? Is there any example to follow?

FezVrasta:
This would also work with Vite

https://vitejs.dev/guide/features.html#web-workers

PR #970, opened 2023-10-12
feat: support dynamic emscripten module import in web workers

thewtex:
Firefox now has dynamic ESM support in web workers, and it is enabled in the current stable release.

This means we can:

1. Allow WebPack, Vite, Rollup, to bundle the web workers as usual by default and they can use the dynamic import at runtime.
2. Not build the UMD Emscripten module wrapper for importScript imports

This patch sets the bundler handling of pipeline.worker.js to be the default, and stops building the UMD wasm modules.


thewtex:
Addressing #476 

@jadh4v @PaulHax

src/itk-wasm-cli.js
```diff
@@ -250,9 +250,9 @@ function bindgen(options) {
     if (err.code !== 'EE XIST') throw err
   }
 
-  // Building for emscripten can generate duplicate .umd.wasm and .wasm binaries
+  // Building for emscripten can generate duplicate and .wasm binaries
```
PaulHax, line 253: generate duplicate !!! and .wasm   🤔

thewtex: Good catch :-)

PR #992, opened 2023-11-10
docs: honor nifti case in supported image file format list

jhlegarreta:
Honor the `NIfTI` format case in the supported file format list.


jhlegarreta:
Test failure is unrelated.

thewtex (approved):
@jhlegarreta thank you!

PR #996, opened 2023-11-11
docs: sort image file format list alphabetically

jhlegarreta:
Sort image file format list alphabetically.


thewtex (approved):
:100: @jhlegarreta thank you!

PR #997, opened 2023-11-11
docs: prefer file format acronyms only

jhlegarreta:
- docs: prefer using only acronyms in supported file format list
- docs: remove redundant/unnecessary information in mesh file format names


thewtex (approved):

thewtex:
@jhlegarreta thanks!

Issue #1018, opened 2023-12-06
image-io writeImage API suggetions


PaulHax:
I was surprised a couple times while integrating `writeImage` from @itk-wasm/image-io into VolView here: https://github.com/Kitware/VolView/pull/527

Surprises for the `writeImage` func:
1. After running writeImage with the labelmap as input, painting on the labelmap gave an error ` Cannot perform Construct on a detached ArrayBuffer`.  Would be nice to avoid remembering to `copyImage`.  
2. The first a arguemnt to `writeImage`, is the web worker handle, which could be null.  Feels more like an optional argument, but is has the prominant first parameter position.
3. I had a memory leak bug where I forgot to terminate the webWorker.

Ideas:
- Don't include input image data array in webworker's postMessage transferable arg by default.  Could add `imageArrayTransferable` flag in `options`?
- Hide webworker implimentation detail a little.  Put `inWebWorkerHandle` in `options`.  Without that option, automaticly terminate webworker.
- writeImage could just return the serialzied image, rather than a results object. If couldWrite is going to be false, just return `undefined` as output of writeImage.  Make a utility function that creates a webworker if users want to control lifecycle of webworker or even "pre-cache" the webworker.

thewtex:
Also discussed:

- Function to create the worker beforehand.
- Passing in all assets / states as options.

thewtex:
Pass in wasm binary asset?

thewtex:
Excellent suggestions :+1: 

We now:

- Have the `webWorker` passed optionally on the last `options` argument #1032 
- Do not need to `terminate()` the web worker, but it is still possible to do so. #1032
- Will copy ArrayBuffer's as needed by default, when SharedArrayBuffer is not available, for all interface types. If it is known that the input will not be needed, a `noCopy: true` option can be passed.
- Have a `createWebWorker` to create an explicitly managed web worker beforehand, #1033.

I am not sure if there is a usage case for passing the wasm binary asset, but we can address that in another issue / PR if it comes to bear.

PR #1032, opened 2023-12-22
refactor: webWorker is optionally passed on last argument

thewtex:
Re #1018


PaulHax (commented):
This API change makes sense to me.  In the future, are you thinking of automaticly terminating the worker in `runPipeline` if no worker option is passed?

thewtex:
I added a commit so that calling `.terminate()` is not necessary to avoid resource leaks by default. While `.terminate()` still can optionally be called to release memory, the default behavior is to re-use the worker, which is a better performance / memory usage tradeoff in most use cases. Creating the worker is not that expensive, but instantiating the wasm module and allocating the memory it uses is expensive.

PR #1033, opened 2023-12-22
feat: provide createWebWorker

thewtex:
So compatible web workers can be created explicitly when desired.

### DIFF
--- a/packages/core/typescript/itk-wasm/cypress/e2e/pipeline/run-pipeline.cy.ts
+++ b/packages/core/typescript/itk-wasm/cypress/e2e/pipeline/run-pipeline.cy.ts
@@ -47,7 +47,7 @@ describe('runPipeline', () => {
     })
   })
 
-  it('fetches Wasm files from a custom pipelineWorkerUrl string', () => {
+  it('fetches the pipeline web worker from a custom pipelineWorkerUrl string', () => {
     cy.window().then(async (win) => {
       const itk = win.itk
       const pipelineWorkerUrl = new URL('/itk-wasm-pipeline.worker.js', demoServer).href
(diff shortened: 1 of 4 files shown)


thewtex:
Re: #1018

packages/core/typescript/itk-wasm/cypress/e2e/pipeline/run-pipeline.cy.ts
```diff
@@ -60,6 +60,20 @@ describe('runPipeline', () => {
...
+    cy.window().then(async (win) => {
+      const itk = win.itk
+      const pipelineWorkerUrl = new URL('/itk-wasm-pipeline.worker.js', demoServer).href
+      const webWorker = await itk.createWebWorker(pipelineWorkerUrl)
+
+      const args = []
+      const outputs = null
+      const inputs = null
+      const stdoutStderrPath = 'stdout-stderr-test'
+      const { returnValue, stdout, stderr } = await itk.runPipeline(webWorker, stdoutStderrPath, args, outputs, inputs, { pipelineWorkerUrl })
```
PaulHax, line 73: Is `{ pipelineWorkerUrl }` option still needed?

PaulHax (approved):
Cool.  A step to removing this benign error causing "tee up" of a pipeline in VolView:
https://github.com/Kitware/VolView/blob/main/src/io/dicom.ts#L69-L79

To "prime the pump" for the `.wasm` asset, should VolView just `fetch()` its custom .wasm file?  Is there another possible speedup by asking the browser to "load" the .wasm?

packages/core/typescript/itk-wasm/cypress/e2e/pipeline/run-pipeline.cy.ts
thewtex: Nope! Good catch.

thewtex:
> To "prime the pump" for the .wasm asset, should VolView just fetch() its custom .wasm file? Is there another possible speedup by asking the browser to "load" the .wasm?

I made a note to investigate this more in #1018

PR #1034, opened 2024-01-02
fix(loadPipelineModule): load pipeline once

floryst:
Presently, calling `loadPipelineModule` multiple times with the same pipeline in quick succession will cause multiple network requests to go out to fetch the pipeline code. This change coalesces multiple calls into a single pipeline fetch.


thewtex (approved):
Good catch, thanks @floryst !

thewtex:
Failing tests unrelated.

PR #1058, opened 2024-02-07
feat: add itkwasm-downsample-cucim

thewtex:
This is based on the itk_cucim package work. _discrete_gaussian.py by Greg Lee is copied directly.

We are currently getting sporatic differences in the test output of 1.


thewtex:
@grlee77 @gigony please take a look :pray: . I see `affine_transform` in cucim: https://github.com/rapidsai/cucim/blob/d4ec01eb53afceaaeec36856e1dcbf2c282659d0/python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py#L344-L356 but how to import it? Which is better to use -- that or the one in cupy?

The motivation is to provide NVIDIA acceleration when available for [OME-Zarr generation](https://ngff.openmicroscopy.org/latest/) per https://github.com/thewtex/ngff-zarr/pull/68

thewtex:
@tbirdso @agirault @jjomier per our discussion at the hackathon, we can use GPU representations on the `image.data` pixel buffer. In this case, it is a cupy array. This will reside on the GPU until needed in the CPU memory. In the test here, that is happening in `write_image`.

thewtex:
Failing CI is unrelated.

Issue #1087, opened 2024-03-05
Support for cache refresh of updated pipeline files


pritamrungta:
Hi! I'm not fully sure if this is already present, but here's my use-case

I have vite app with all pipelines located in public/itk/pipelines/*[.wasm|.umd.js|...]
Whenever any changes are made in the pipelines, I would like the browser to fetch the updated file (on execution of `runPipeline`) instead of serving from disk cache.

So, I'm imagining the URLs to be something like `https://hostname/itk/pipelines/filter.wasm?_=<app-version>` instead of plain `https://hostname/itk/pipelines/filter.wasm`

Is there a way to configure the query params like the above or any other better alternative to solve the problem?

Thanks!

thewtex:
Hi!

Makes sense.

Perhaps we add a `setPipelinesQueryParams` to specify query params that are appended to pipeline asset requests for cache busting?

pritamrungta:
Yes, that works.

(I'm also curious if you encountered the issue more often and if so, what's your general approach to overcome that)

thewtex:
Could you please [create a patch](https://wasm.itk.org/en/latest/development/hacking_itk_wasm.html)?

For production deploys, I often use a service worker, where there is more direct control over the cache. But we should have this feature in ITK-Wasm.

pritamrungta:
Sure, will do.

thewtex:
[Released in 1.0.0-b.170](https://github.com/InsightSoftwareConsortium/itk-wasm/releases/tag/itk-wasm-v1.0.0-b.170).

PR #1093, opened 2024-03-11
feat(itk-wasm): #1087 - Add query params options to pipeline URL

pritamrungta:
Add pipeline URL query params options to support cache busting of updated assets


thewtex (approved):
@pritamrungta most excellent! :1st_place_medal: 

Failing tests are unrelated.

PR #1096, opened 2024-03-11
fix(wasi-function-module): use file path variable name

PaulHax:
Getting runtime type error in `path` with 
```
        for value in files:
            input_file = str(PurePosixPath(files))
``` 
Changes to 
```
        for value in files:
            input_file = str(PurePosixPath(value))
```

### DIFF
--- a/packages/core/typescript/itk-wasm/src/bindgen/python/wasi/wasi-function-module.js
+++ b/packages/core/typescript/itk-wasm/src/bindgen/python/wasi/wasi-function-module.js
@@ -180,7 +180,7 @@ from itkwasm import (
         const interfaceType = interfaceJsonTypeToInterfaceType.get(parameter.type)
         if (interfaceType.includes('File')) {
           // for files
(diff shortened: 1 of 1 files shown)


packages/core/typescript/itk-wasm/src/bindgen/python/wasi/wasi-function-module.js
```diff
@@ -180,7 +180,7 @@ from itkwasm import (
         const interfaceType = interfaceJsonTypeToInterfaceType.get(parameter.type)
         if (interfaceType.includes('File')) {
           // for files
-          args += `            input_file = str(PurePosixPath(${snakeCase(parameter.name)}))\n`
+          args += `            input_file = str(PurePosixPath(value))\n`
```
thewtex, line 183: Looks reasonable

PaulHax: Otay!

thewtex (approved):

PR #1105, opened 2024-04-10
fix(create-web-worker): allow undefined workerUrl

floryst:
Calling `createWebWorker()` with an undefined `pipelineWorkerUrl` should be equivalent to passing in `null`.


thewtex (approved):
@floryst thanks!

Other CI issues should be unrelated.

PR #1172, opened 2024-07-04
Fix itkPipeline help stream parsing crash and native msvc static debug build

jadh4v:
# fix(itkPipeline): parameters parsing crash when no space in line 

Fix a bug where a call to --help and printing of parameters for
a cli app would crash if a description or parameter line
did not contain any space.

# build(msvc): add cmake flag to enable msvc static runtime debug library 

ITK-Wasm prefers static build. For native build on windows (msvc),
it is necessary to enable/specify the use of static runtime debug
library indicated by the /MTd compiler flag.
Add a CMAKE flag (default: ON) to enable/disable this compiler
flag when msvc is being used for building.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,20 @@ endif()
 
 set(WebAssemblyInterface_LIBRARIES WebAssemblyInterface)
 
+if(MSVC) #-- Configure MSVC_STATIC_RUNTIME only if using MSVC environment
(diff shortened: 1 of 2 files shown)


thewtex (approved):
@jadh4v yes, looks good, thanks!

https://github.com/InsightSoftwareConsortium/ITK-Wasm/actions/runs/9796643484/job/27051541857?pr=1172

PR #1180, opened 2024-07-16
fix(OutputImage): dataSize needs components per pixel

thewtex:
As a follow-up to f9f2805766ae1da281dc35dc6c1b3405911007f3, we still
need the component per pixel when computing the data size.

### DIFF
--- a/include/itkOutputImage.h
+++ b/include/itkOutputImage.h
@@ -84,7 +84,7 @@ public:
 
         const auto dataAddress = reinterpret_cast< size_t >( wasmImage->GetImage()->GetBufferPointer() );
         using ConvertPixelTraits = DefaultConvertPixelTraits<typename ImageType::PixelType>;
-        const auto dataSize = wasmImage->GetImage()->GetPixelContainer()->Size() * sizeof(typename ConvertPixelTraits::ComponentType);
+        const auto dataSize = wasmImage->GetImage()->GetPixelContainer()->Size() * sizeof(typename ConvertPixelTraits::ComponentType) * ConvertPixelTraits::GetNumberOfComponents();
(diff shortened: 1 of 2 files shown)


include/itkOutputImage.h
```diff
@@ -84,7 +84,7 @@ class ITK_TEMPLATE_EXPORT OutputImage
 
         const auto dataAddress = reinterpret_cast< size_t >( wasmImage->GetImage()->GetBufferPointer() );
         using ConvertPixelTraits = DefaultConvertPixelTraits<typename ImageType::PixelType>;
-        const auto dataSize = wasmImage->GetImage()->GetPixelContainer()->Size() * sizeof(typename ConvertPixelTraits::ComponentType);
+        const auto dataSize = wasmImage->GetImage()->GetPixelContainer()->Size() * sizeof(typename ConvertPixelTraits::ComponentType) * wasmImage->GetImage()->GetNumberOfComponentsPerPixel();
```
jadh4v, line 87: I believe, the reason I removed the "NumberOfComponents" part is because the call to `PixelContainer->Size()` takes into account the number of components for a `VectorImage`. PixelContainer is constructed on `InternalPixelType` and does not appear to be aware of `numberOfComponents`.
See: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkVectorImage.h#L144

thewtex: I do not think `PixelContainer->Size()` accounts for the components. The definition is the same for `Image`: https://github.com/InsightSoftwareConsortium/ITK/blob/f88a2896a7449088af0a29070e53043f85ad6a3e/Modules/Core/Common/include/itkImage.h#L145

jadh4v (approved):
LGTM

PR #1188, opened 2024-07-26
Add quay prefix to docker image names

jadh4v:
Prefix docker image names with registry paths so that they work with `podman` without setting local environment for unqualified-search-registries.


jadh4v:
Not sure why some of the wasm binaries get modified by the build. Do we want to check-in these files?


image

thewtex (approved):
:+1:

PR #1190, opened 2024-08-02
fix(runPipelineNode): unmounting overlapping paths

jadh4v:
If multiple paths are mounted such that one is a parent of the other, then you may encounter the following exceptions:

```
Rejected promise returned by test. Reason:

  ErrnoError {
    code: 'EINVAL',
    errno: 28,
    message: 'Invalid argument',
  }

Rejected promise returned by test. Reason:

  ErrnoError {
    code: 'ENOENT',
    errno: 44,
    message: 'No such file or directory',
  }
```

This fix will determine paths that are safely unmountable, and only unmount each path once.


thewtex (approved):
:+1:

PR #1193, opened 2024-08-07
Fix and add tests for computation of buffer size when converting VectorImage between native-image and wasm-image.

jadh4v:


thewtex (approved):
Looks good, thanks.

CI failures are due to unrelated refactoring in #1191

PR #1209, opened 2024-08-23
native itk dcmtk bumps

thewtex:
- ci(python-wasm): disable dicom tests
- chore(native): bump ITK, DCMTK to match Docker
- build(native): add patch to fix ITKTotalVariation build on Linux
- build: re-use ITK, DCMTK source locations in native, docker builds


itk_wasm_env.bash
```diff
@@ -5,16 +5,16 @@ function die() {
     exit 1
 }
 
-if [$OSTYPE == "cygwin"] || [$OSTYPE == "msys"] || [$OSTYPE =="win32"]; then
+if test "$OSTYPE" = "cygwin" || test "$OSTYPE" = "msys" || test "$OSTYPE" = "win32"; then
     echo "Windows platform detected ... adding \"/Zc:__cplusplus /DNOMINMAX\" to \$CXXFLAGS"
     export CXXFLAGS="/Zc:__cplusplus /DNOMINMAX"
```
jadh4v, line 10: My current script shows, but can't remember why I updated it. I will figure it out and post here again:
```bash
export CXXFLAGS="/Zc:__cplusplus /Zc:preprocessor /DNOMINMAX"
```

.gitignore
```diff
@@ -102,4 +102,4 @@ packages/transform-io/typescript/demo-app/
 *.egg-info
 
 native/
-
+src/docker/itk-wasm-base/itk_wasm_env_vars.sh
```
jadh4v, line 105: why gitignore though? we don't expect commits often but expect local changes to the file?

itk_wasm_env.bash
jadh4v: Or I can just update it in a different PR if `/Zc:preprocessor` is a legit requirement for MSVC native builds.

jadh4v (approved):
This is great. Let me know if you'd like me to try the build on native windows.
LGTM.

itk_wasm_env.bash
thewtex: @jadh4v thanks! I added that flag.

I am going to merge to rebase a different branch on this, but if you test and find any issues with MSVC, another patch would be appreciated.

.gitignore
thewtex: It is a derived file from ITK-Wasm/itk_wasm_env.bash. We need to add in there since Docker's build context only takes into account the local directory.

PR #1260, opened 2024-10-29
fix(include): avoid segfault when calculating totalSize

kgoettler:
Hi,

This PR fixes an issue in which a segfault could occur when calculating the total size of the buffered region. If my understanding is correct here, `imageJSON.size[i]` should be used if `imageJSON.bufferedRegion.size` is undefined.

Related thread on the ITK forum: https://discourse.itk.org/t/running-pipeline-inputs-outputs-tutorial-in-browser/7282


thewtex (approved):
👏 @kgoettler thank you for the contribution!

PR #1264, opened 2024-10-31
feat(itk-wasm-cli): Update default Docker image for 20241030-6c93a1be

thewtex:


thewtex:
@kgottler pushed your patch in #1260 to DockerHub and quay.io

kgoettler:
@thewtex Excellent - thank you very much!

Issue #1289, opened 2024-12-06
Build relies on OS install of python and fails if it's not there


pieper:
This issue is broken out from #1281 

@thewtex this is where the build fails because it cannot find python.

I agree with your comment that it would make sense for ITK-Wasm to use pixi for all python steps in order to have some control over versioning.

The issue is in dcmqi, so @fedorov may also need to be involved.

```
? Pixi task (configure-native-dicom in native): cmake -B$ITK_WASM_NATIVE_WORKSPACE/dicom-build -Spackages/dicom -GNinja
  -DITK_DIR:PATH=$ITK_WASM_ITK_BUILD_DIR
  -DBUILD_TESTING:BOOL=ON
  -DCMAKE_CXX_STANDARD:STRING=20
  -DCMAKE_BUILD_TYPE:STRING=Debug: (Configure dicom native binaries)
-- The C compiler identification is GNU 12.4.0
-- The CXX compiler identification is GNU 12.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /home/exouser/ITK-Wasm/.pixi/envs/native/bin/x86_64-conda-linux-gnu-cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/exouser/ITK-Wasm/.pixi/envs/native/bin/x86_64-conda-linux-gnu-c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found OpenMP_C: -fopenmp (found version "4.5")
-- Found OpenMP_CXX: -fopenmp (found version "4.5")
-- Found OpenMP: TRUE (found version "4.5")
-- Setting C++ standard
-- Setting C++ standard - C++20
-- Configuring with build type 'Debug'
-- Checking if building as a Slicer extension
-- Checking if building as a Slicer extension - no (DCMQI_BUILD_SLICER_EXTENSION is OFF)
-- Checking if using built-in JsonCpp
-- Checking if using built-in JsonCpp - yes
CMake Warning (dev) at /home/exouser/ITK-Wasm/native/dicom-build/_deps/dcmqi_lib-src/CMakeLists.txt:251 (find_package):
  Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
[...]

PR #1290, opened 2024-12-10
build(dicom): ensure dcmqi finds pixi's Python

thewtex:
Through upstream patch:

  https://github.com/QIICR/dcmqi/pull/511


thewtex:
To address #1289

PR #1420, opened 2025-07-17
docs: update C++ installation / overview

thewtex:


thewtex:
@N-Dekker how does this look?

PR #1422, opened 2025-07-18
STYLE: Declare data of Stream classes `private`, instead of `protected`

N-Dekker:
Following C++ Core Guidelines, Jul 8, 2025, ["Avoid `protected` data"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-protected)


thewtex:
:+1: @N-Dekker thank you!

> type must be lower-case [type-case]
   type must be one of [build, chore, ci, docs, feat, fix, perf, refactor, revert, style, test] [type-enum]
   
These are the conventions of a broadly adopted standard, Conventional Commit:

  https://www.conventionalcommits.org/en/v1.0.0/
  
that is used throughout the web community, and has a bunch of tooling built around it. Since this is a web-oriented project and we use some of that tooling, we use this convention.

thewtex (approved):
🙏

PR #1426, opened 2025-07-21
Add InputStreamBase: common base class of InputBinaryStream and InputTextStream

N-Dekker:


include/itkInputBinaryStream.h
```diff
@@ -40,66 +32,18 @@ namespace wasm
...
  * \brief Input binary std::istream for an itk::wasm::Pipeline
  *
- * This stream is read from the filesystem or memory when ITK_WASM_PARSE_ARGS is called.
- *
- * Call `Get()` to get the std::istream & to use an input to a pipeline.
- *
  * \ingroup WebAssemblyInterface
  */
-class WebAssemblyInterface_EXPORT InputBinaryStream
+class InputBinaryStream : public InputStreamBase
```
thewtex, line 37: We still need `WebAssemblyInterface_EXPORT`

thewtex (changes requested):
@N-Dekker thanks!

One issue noted inline

include/itkInputBinaryStream.h
N-Dekker: > We still need `WebAssemblyInterface_EXPORT`

The `WebAssemblyInterface_EXPORT` on InputBinaryStream and InputTextStream caused link errors, at https://open.cdash.org/viewBuildError.php?buildid=10550428, like:

> itkPipelineTest.cxx.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __cdecl itk::wasm::InputTextStream::InputTextStream(void)"

Removing those two `WebAssemblyInterface_EXPORT` fixed the compilation! With the very last [force-push](https://github.com/InsightSoftwareConsortium/ITK-Wasm/compare/3a08ecd038180af32a2b21cf5f8c890e3b52f2d1..6aac66e2d216d2d9e27addcf7362fe22a719a6d4). You see, the CI is finally happy with this PR!

With this PR, everything that is "DLL-exported" is moved to the new base class, `InputStreamBase`, which has a cxx file. `InputStreamBase` _does_ have  `WebAssemblyInterface_EXPORT`. On the other hand, InputBinaryStream and InputTextStream no longer have a cxx file, they are entirely _inline_. So these two derived classes do not need  `WebAssemblyInterface_EXPORT`, technically. They don't even compile with `WebAssemblyInterface_EXPORT`!

Do you still want InputBinaryStream and InputTextStream to have `WebAssemblyInterface_EXPORT`? If so, do you have a suggestion how to make them compile?

thewtex: Great, thanks for the explanation, Niels.

PR #1516, opened 2026-04-29
ENH: Add ITKTestingData gh-pages URL as primary CID source

hjmjohnson:
Add the `ITKTestingData` gh-pages mirror as the primary remote source in `WebAssemblyInterface/CMake/ITKExternalData.cmake`, after the local IPFS gateway and ahead of itk.org / IPFS public gateways.

`%(algo)` substitutes the uppercase algorithm name, matching the case-sensitive `/CID/`, `/MD5/`, `/SHA512/` directory layout on the gh-pages branch (parallel to commit 317ab5fdaed2 on the ITK side).


thewtex (approved):
👍

PR #1545, opened 2026-06-30
fix(OutputImage): support information-only image outputs

vboussot:
## Summary

- allow memory-I/O image outputs with an empty buffered region
- reshape Python image data from `bufferedRegion.size` while preserving the
  largest possible image `size`
- apply the same buffered-region reconstruction in Pyodide
- add native, Python runtime, and Pyodide regression coverage

## Context

`downsample_bin_shrink(..., information_only=True)` calls
`UpdateOutputInformation()` and intentionally produces output metadata without
allocating a pixel buffer. `OutputImage` currently aborts when that legitimate
buffer has zero bytes. After removing the abort, the Python runtime also needs
to reshape data from the buffered region rather than the largest possible
region; otherwise it attempts to reshape an empty array to the logical image
size.

This breaks the metadata pre-pass used by `itkwasm-downsample-cucim==0.2.0`
for Gaussian and bin-shrink downsampling.

Downstream report: https://github.com/fideus-labs/ngff-zarr/issues/577

## Validation

- core Python tests excluding browser-only tests: **28 passed, 2 skipped**
- focused native `itkPipelineMemoryIOTest`: **passed**
- C++ files pass `clang-format --dry-run --Werror`
- full native build reached and compiled the modified test, but the aggregate
  target was blocked by an unavailable pre-existing external test object:
  `TransformSequence.h5`
- WASI package rebuild was not available locally because Docker/Podman is not
  installed
- the Pyodide wheel built successfully, but all browser tests were blocked at
  setup because the local Chrome instance exited before creating a session

### DIFF
--- a/include/itkOutputImage.h
+++ b/include/itkOutputImage.h
@@ -94,11 +94,6 @@ public:
         const auto dataSize = wasmImage->GetImage()->GetBufferedRegion().GetNumberOfPixels() *
                               sizeof(typename ConvertPixelTraits::ComponentType) *
                               wasmImage->GetImage()->GetNumberOfComponentsPerPixel();
-        if (dataSize <= 0)
-        {
-          std::cerr << "dataSize cannot be zero or negative." << std::endl;
-          abort();
-        }
         setMemoryStoreOutputArray(0, index, 0, dataAddress, dataSize);
 
         const auto directionAddress =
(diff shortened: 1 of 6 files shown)


thewtex (approved):
@vboussot 👏

PR #1549, opened 2026-07-03
feat(downsample): add resample-bounding-box pipeline (metadata-only, C++/TS/Python)

thewtex:
## Summary

Adds a new **`resample-bounding-box`** pipeline to the `@itk-wasm/downsample` / `itkwasm-downsample` package. Given a spatial `transform`, a `fixed` image, and a `moving` image, it computes the padded sub-region of the **moving** image that a caller must fetch in order to resample the **fixed** image's grid through that transform — using **image metadata only** (size, spacing, origin, direction). No pixel buffer is ever dereferenced, so both images may be passed with empty `data`.

The pipeline is implemented once in C++ and shipped across all three runtimes — C++ (WASI/native CTest), TypeScript (browser + Node), and Python (WASI + Emscripten) — producing identical results.

## Why

Resampling a fixed grid into a moving image only ever reads moving-image samples inside the transformed footprint of the fixed grid. When the moving image is large, remote, or chunked, materializing all of it just to resample a small overlapping region is wasteful. This pipeline answers *"which moving-image indices will the resample actually read?"* from arithmetic on metadata alone, so a caller can fetch only that block (or skip non-overlapping tiles entirely) before moving any real pixels. It complements the existing resampling pipelines in this package and the affine-transform tooling used to build the transforms fed into it.

## What changed

**New pipeline (C++)**
- `resample-bounding-box.cxx` — pipeline entry point; parses `transform`, `fixed`, `moving`, `bounding-box` (output JSON) and a `--padding` option, and serializes the result.
- `resampleBoundingBox.h` — the metadata-only bounding-box computation.
- `resampleReadInputTransform.h` — generic transform reading/dimension peeking.
- `resample-bounding-box-test.cxx` + `resample-bounding-box-generate-inputs.cxx` — CTest coverage and a self-contained test-input generator.
- `CMakeLists.txt` wired up for the new targets.

**Language bindings & tests**
- **TypeScript**: `resampleBoundingBox` (browser + Node) with generated options/result interfaces, exported from the package index; Node tests and a browser demo-app controller.
- **Python**: `resample_bounding_box` (sync + async) for WASI and Emscripten packages, with WASI pytest coverage.

**Documentation**
- `docs/resample-bounding-box.md` — motivation, inputs, the algorithm, the output JSON schema, edge cases, and intended downstream use.

**Core `itk-wasm` bindgen improvements** (needed by this pipeline, but general fixes)
- **Falsy numeric options are now forwarded.** TypeScript and Python binding generators previously used a truthiness guard, which silently dropped a valid `0` for a numeric option (e.g. `--padding 0`). They now use a presence check (`typeof !== "undefined"` / `is not None`) for numeric options while keeping the truthiness guard for `TEXT`/`BOOL`.
- **New opt-in `itk-wasm.bindgen-exclude` config** in `package.json` so build-only helper executables (here, the `resample-bounding-box-generate-inputs` test-input generator) are not emitted as public language bindings.

## Implementation details

- **Metadata-only contract.** Only size/spacing/origin/direction are read; the fixed and moving pixel buffers are never dereferenced.
- **Dimension dispatch.** Supports transforms at dimension 2, 3, and 4. The pipeline dispatches on the transform dimension *itself* rather than via `itk::wasm::SupportInputTransformTypes`, because that helper deserializes the transform input (a `TransformList` / JSON array) as a single transform object and throws for every in-memory transform. `--help`/`--interface-json`/`--version` dispatch to a default 2D functor so bindgen can extract the interface.
- **Double-precision math.** The transform is read into the abstract `itk::Transform` base at double precision — lossless for float32/float64 inputs and robust to `.iwt` scalar-type detection that can misreport a float64 transform as float32.
- **Full-boundary sampling, not just corners.** Every boundary pixel of the fixed grid (all faces/edges, not just the `2^N` corners) is transformed to a moving-image continuous index; interior pixels are skipped efficiently. For affine transforms the transformed rectangle is convex so corners would suffice, but for nonlinear transforms an interior edge pixel can map outside the corner hull — sampling the full boundary avoids under-bounding the region.
- **Padding & edge cases.** `--padding` (default `1`, covering linear interpolation's one-neighbor read) expands the integer region symmetrically per side; sizes are clamped to ≥ 0; a degenerate fixed image with a zero-length axis yields an empty all-zero region. The unpadded `corners` are the tight transformed-point extremes regardless of padding.

### Output JSON

```json
{
  "paddedStartIndex": [int, ...],
  "paddedSize": [uint, ...],
  "paddedCorners": { "min": [double, ...], "max": [double, ...] },
  "corners":       { "min": [double, ...], "max": [double, ...] }
}
```

## Test plan

- C++ CTest asserts region results and the boundary-pixel count (guarding against a regression to corners-only sampling), covering 2D/3D translation and a 2D affine rotation with non-axis-aligned corners.
- Node and Python tests assert identical results to the C++ pipeline across the same cases, including the metadata-only (empty-buffer) contract and `--padding 0`.


thewtex:
@vboussot please take a look

PR #1550, opened 2026-07-09
Add parameter-driven `resample` pipeline to the downsample package

thewtex:
## Summary

Adds a new parameter-driven `resample` pipeline to the `downsample` package. Unlike the existing `resample-to-reference` pipeline — which copies an output grid's geometry from a supplied reference image — `resample` lets callers specify the target grid **explicitly** via parameters, with each parameter defaulting per-axis from the input (moving) image when omitted. This makes common operations (change spacing, crop/pad, reorient, resample onto an arbitrary grid) possible without having to construct a reference image first.

## Why

`resample-to-reference` requires a full reference image just to describe the desired output geometry. That is awkward when you only want to, say, halve the spacing or set an explicit size — you have to fabricate an image with the right metadata. The new `resample` pipeline exposes the output grid directly as parameters so no reference image is needed.

## What changed

**New C++ pipeline** — `packages/downsample/resample.cxx`, a close clone of `resample-to-reference.cxx`. It reuses the existing infrastructure verbatim: the shared generic transform reader (`resampleReadInputTransform.h`), the `SelectInterpolator` six-name helper, the default/background-value option, and both the scalar and `itk::VectorImage` code paths. The only genuinely new logic is **output-grid resolution with input-geometry fallback**.

**Parameter surface** (all optional, each defaulting per-axis from the moving image):

| Parameter | Meaning |
|---|---|
| `size` | Output size in pixels per axis |
| `outputSpacing` | Output spacing per axis, physical units |
| `outputOrigin` | Physical coordinates of the first pixel |
| `outputDirection` | D×D orientation matrix, flattened row-major |
| `transform` | Transform mapping output points into moving-image space (composite semantics; defaults to identity) |
| `interpolator` | Interpolation method |
| `defaultValue` | Background value for samples mapping outside the moving image (default `0`) |

Notably there is **no `referenceImage` parameter**.

**Generated bindings** — Built the emscripten (`resample.wasm`) and WASI (`resample.wasi.wasm`) targets and regenerated the TypeScript and Python bindings from the compiled wasm: `resample{,-node}{,-options,-result}.ts`, the appended `resample`/`resampleNode` exports, the top-level/WASI/emscripten Python `resample` modules with `__init__` re-exports, the re-embedded base64 worker bundle, and the demo `resample-controller.ts` + `sl-tab`. Wasm binaries themselves are gitignored; only generated source and the embedded worker bundle land in git.

**Tests** — Six self-contained C++ CTests plus Node (ava) and Python (wasi pytest) integration tests through the in-memory wasm path, covering: spacing-only auto-size geometry, explicit-full-grid equivalence to `resample-to-reference`, full-default identity reproduction, `VectorImage` identity reproduction, and the out-of-bounds background value. No new test data or baselines are introduced — tests validate against `resample-to-reference` (equivalence), the input image itself (identity), or directly-asserted geometry.

## Implementation details

- **Auto-size**: when `outputSpacing` is provided without `size`, the size auto-computes as `ceil(inputSize * inputSpacing / spacing)` per axis (clamped to `>= 1`) so the input's physical extent is preserved at the new spacing.
- **Validation**: provided vector lengths are validated against the image dimension and reported via `pipeline.exit()`.
- The `VectorImage` identity test normalizes `apple.jpg`'s fractional spacing to a unit grid so the identity reproduction is bit-exact.
- The pipeline links cleanly under both the WASI and emscripten toolchains; all six WASI CTests pass end-to-end.


thewtex:
@vboussot @sedghi the other resample pipeline we discussed

vboussot:
Nicely done, this covers the needs I had in mind now. Like that --size defaults
from --output-spacing too.

PR #1581, opened 2026-08-06
fix: set transform parameters by value when reconstructing from JSON

### LINKED ISSUE #1580: resample-bounding-box aborts in WebAssembly for any BSplineTransform
# resample-bounding-box aborts in WebAssembly for any BSplineTransform

`resample-bounding-box` aborts with no diagnostic (`wasm trap: wasm 'unreachable'
instruction executed`, empty stderr) for **any** `BSplineTransform`. Every other
parameterization I tested works, affine, Euler2D/3D, Similarity2D,
VersorRigid3D, and a **non-linear displacement field**, so this is specific to
B-splines, not to non-linear transforms.

It blocks the deformable use case: a rigid -> affine -> bspline Elastix
registration (the parameter maps used in ITKElastix#375) produces a composite
that aborts; removing its B-spline component makes the same call succeed.

**Environment:** itkwasm-downsample 2.0.0, itkwasm 1.0b200, itk 5.4.6, Python 3.14, Linux x86_64.

## Reproduction

`pip install "itkwasm-downsample==2.0.0"`:

```python
import numpy as np
from itkwasm import (FloatTypes, Image, ImageType, IntTypes, PixelTypes,
                     Transform, TransformParameterizations, TransformType)
from itkwasm_downsample import resample_bounding_box

def image(size):
    return Image(imageType=ImageType(dimension=2, componentType=IntTypes.UInt8,
                                     pixelType=PixelTypes.Scalar, components=1),
                 origin=[0.0, 0.0], spacing=[1.0, 1.0], direction=np.eye(2),
                 size=size, data=np.empty((0,), dtype=np.uint8))

def transform(parameterization, fixed_params, params):
    return Transform(
        transformType=TransformType(transformParameterization=parameterization,
                                    parametersValueType=FloatTypes.Float64,
                                    inputDimension=2, outputDimension=2),
        numberOfFixedParameters=len(fixed_params), numberOfParameters=len(params),
        fixedParameters=np.asarray(fixed_params, dtype=np.float64),
        parameters=np.asarray(params, dtype=np.float64))

fixed, moving = image([64, 64]), image([256, 256])

# Control: affine -> OK
affine = transform(TransformParameterizations.Affine,
                   [0.0, 0.0], [1.0, 0.0, 0.0, 1.0, 5.0, 3.0])
print(resample_bounding_box([affine], fixed, moving, padding=1))

# 4x4 cubic mesh over [0,64]^2 -> 7x7 grid, 98 coefficients, ALL ZERO
# (mathematically the identity) -> aborts
bspline = transform(TransformParameterizations.BSpline,
                    [7.0, 7.0, -16.0, -16.0, 16.0, 16.0, 1.0, 0.0, 0.0, 1.0],
                    np.zeros(98))
print(resample_bounding_box([bspline], fixed, moving, padding=1))
```

Affine returns `{'paddedStartIndex': [4, 2], 'paddedSize': [66, 66], ...}`;
the B-spline raises `wasmtime._trap.Trap`. BSpline aborts in both 2D and 3D.

Since an all-zero B-spline aborts too, the coefficient values are irrelevant.
The input is sound: its `dict_from_transform` output round-trips through
`itk.transform_from_dict` to the identical mapping.

## Root cause

`reconstructTransformEntry` in `packages/downsample/resampleReadInputTransform.h`
fills the transform with `CopyInParameters`, which does:

```cpp
std::copy(begin, end, this->m_Parameters.data_block());   // no resize
```

The base class documents the precondition: *"The range of values must conform to
`std::copy(begin, end, m_Parameters)` requirements"*. That holds for transforms
sizing `m_Parameters` in their constructor, but not for `itk::BSplineTransform`:
its parameter count is unknown until the grid is set, and it keeps coefficients
in its own `m_InternalParametersBuffer` (`GetParameters()` returns that buffer,
not the base `m_Parameters`, which is never sized). The copy runs past the end of
a zero-length buffer — natively, the same call gives a **SIGSEGV**.

## Proposed fix

Use the by-value setters  `SetParametersByValue` exists precisely for transforms
that would otherwise keep a reference:

```diff
+#include <algorithm>

   if (transformJSON.numberOfFixedParameters > 0)
   {
-    transform->CopyInFixedParameters(fixedParametersPtr,
-                                     fixedParametersPtr + transformJSON.numberOfFixedParameters);
+    FixedParametersType fixedParameters(transformJSON.numberOfFixedParameters);
+    std::copy(fixedParametersPtr,
+              fixedParametersPtr + transformJSON.numberOfFixedParameters,
+              fixedParameters.data_block());
+    transform->SetFixedParameters(fixedParameters);
   }

   if (numberOfParameters > 0)
   {
-    std::vector<double> parameters(numberOfParameters);
+    ParametersType parameters(numberOfParameters);
     ...
-    transform->CopyInParameters(parameters.data(), parameters.data() + numberOfParameters);
+    transform->SetParametersByValue(parameters);
   }
```

Fixed parameters must still be set first: for a B-spline they define the grid,
and hence the number of parameters.

**Verified.** I rebuilt `packages/downsample` for WASI with this patch and re-ran
the reproduction against the rebuilt module, swapping only the `.wasm` file:

```
published module   controls 3/3   b-splines 0/3
patched module     controls 3/3   b-splines 3/3
```

The result is correct, not merely non-crashing. For a B-spline whose boundary
bulges past the hull of its transformed corners — where a corners-only bound
would be wrong — the reported corners match a brute-force full-boundary walk
exactly, and the region is `floor`/`ceil` of it at padding 0, 1 and 3. A real
Elastix rigid → affine → bspline composite also matches a brute-force walk to
the last digit. The upstream `ctest` suite is **25/25** on the patched build.

`include/itkWasmTransformToTransformFilter.hxx` uses the identical pattern at
lines 231/234 and 244/247 — untested, but likely the same defect on the generic
path.

vboussot:
CopyInFixedParameters/CopyInParameters std::copy into the transform's existing m_FixedParameters/m_Parameters without resizing them; the base class documents the requirement as "the range of values must conform to std::copy(begin, end, m_Parameters)". That holds for transforms that size those members in their constructor, but not for itk::BSplineTransform: its parameter count is unknown until the grid is set, and it keeps its coefficients in its own m_InternalParametersBuffer, so the base m_Parameters stays empty. Copying into it runs off the end of a zero-length buffer, which segfaults natively and aborts the module with no diagnostic under WebAssembly.

Use SetFixedParameters and SetParametersByValue instead, which every transform implements as a copy, and check the parameter count so a mismatch reports instead of corrupting memory. Fixed parameters are set first: for a B-spline they define the grid, and hence the number of parameters.

Fixed in both places that reconstruct a transform from TransformListJSON: itkWasmTransformToTransformFilter, in its composite and non-composite branches, and the copy of that pattern in the downsample package's resample pipelines.

Adds itkWasmTransformInterfaceBSplineTest, which round-trips a composite containing a B-spline through the interface. No existing test exercised a B-spline on this path, which is why the defect went unnoticed; the new test segfaults without this change.

Closes #1580

### DIFF
--- a/include/itkWasmTransformToTransformFilter.hxx
+++ b/include/itkWasmTransformToTransformFilter.hxx
@@ -21,6 +21,7 @@
 #include "itkTransformFactoryBase.h"
 
 #include "itktransformParameterizationString.h"
+#include <algorithm>
 #include <exception>
 #include "itkWasmMapComponentType.h"
 #include "itkWasmMapPixelType.h"
@@ -226,12 +227,30 @@ WasmTransformToTransformFilter<TTransform>::GenerateData()
       // Correct extra reference count from CreateInstance()
       ptr->UnRegister();
 
-      FixedParametersValueType * fixedPtr = reinterpret_cast<FixedParametersValueType *>(
+      // CopyIn*Parameters std::copy into m_FixedParameters/m_Parameters without resizing them.
+      // itk::BSplineTransform never sizes the base m_Parameters, so that overruns a zero-length
+      // buffer. Fixed parameters are set first: for a B-spline they define the grid, and hence
+      // the number of parameters.
+      const FixedParametersValueType * fixedPtr = reinterpret_cast<const FixedParametersValueType *>(
         std::strtoull(transformJSON.fixedParameters.substr(35).c_str(), nullptr, 10));
-      ptr->CopyInFixedParameters(fixedPtr, fixedPtr + transformJSON.numberOfFixedParameters);
-      ParametersValueType * paramsPtr = reinterpret_cast<ParametersValueType *>(
+      typename ComponentTransformType::FixedParametersType componentFixedParameters(
+        transformJSON.numberOfFixedParameters);
+      std::copy(fixedPtr,
+                fixedPtr + transformJSON.numberOfFixedParameters,
+                componentFixedParameters.data_block());
+      ptr->SetFixedParameters(componentFixedParameters);
+
+      const ParametersValueType * paramsPtr = reinterpret_cast<const ParametersValueType *>(
         std::strtoull(transformJSON.parameters.substr(35).c_str(), nullptr, 10));
-      ptr->CopyInParameters(paramsPtr, paramsPtr + transformJSON.numberOfParameters);
+      typename ComponentTransformType::ParametersType componentParameters(transformJSON.numberOfParameters);
+      std::copy(paramsPtr, paramsPtr + transformJSON.numberOfParameters, componentParameters.data_block());
+      if (componentParameters.Size() != ptr->GetNumberOfParameters())
+      {
+        itkExceptionMacro("The input transform carries " << componentParameters.Size() << " parameters but "
+                                                         << transformType << " expects "
+                                                         << ptr->GetNumberOfParameters() << '.');
+      }
+      ptr->SetParametersByValue(componentParameters);
 
       using CompositeTransformType = CompositeTransform<ParametersValueType, TransformType::InputSpaceDimension>;
       CompositeTransformType * compositeTransform = dynamic_cast<CompositeTransformType *>(transform);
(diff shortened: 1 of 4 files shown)


thewtex (approved):
🥇

